PR #1, opened 2021-02-24
Mandatorycontent progress

amit-kumaryadav:


src/main/resources/application.properties
```diff
@@ -33,4 +33,5 @@ spring.data.cassandra.sb.username=cassandra
 spring.data.cassandra.sb.password=cassandra
 
 sunbird.course.service.host=https://staging.open-sunbird.org/api/course/
```
karthik-tarento, line 35: The value should contain - http://lms-service:9000

karthik-tarento (changes requested):

src/main/resources/application.properties
amit-kumaryadav: Pushed the changes

PR #14, opened 2021-06-01
Workallocation v2

amit-kumaryadav:


src/main/java/org/sunbird/workallocation/service/AllocationServiceV2.java
```diff
@@ -52,7 +52,7 @@
     @Value("${workorder.index.type}")
     public String workOrderIndexType;
 
-    @Value("${workallocation.index.name}")
+    @Value("${workallocationv2.index.name}")
```
karthik-tarento, line 55: you suppose to create new --- instead of replacing previous one ?

karthik-tarento (approved):
There is one comment... update the same in the main branch

src/main/java/org/sunbird/workallocation/service/AllocationServiceV2.java
amit-kumaryadav: We have created the new only. It's allocationservicev2

PR #25, opened 2021-10-01
competency Search API integrated also New Id has created for profile Redirection 

Jenifer-Ramsingh:


src/main/java/org/sunbird/searchby/controller/SearchByController.java
```diff
@@ -3,13 +3,15 @@
...
+import org.springframework.web.bind.annotation.*;
+import org.sunbird.assessment.dto.AssessmentSubmissionDTO;
+import org.sunbird.searchby.dto.SearchByFilter;
 import org.sunbird.searchby.service.SearchByService;
 
+import javax.validation.Valid;
+
 @RestController
-public class SearchByController {
+public class SearchByController<requestBody> {
```
karthik-tarento, line 14: no need to provide the <requestBody> in class

src/main/java/org/sunbird/searchby/service/SearchByService.java
```diff
@@ -42,10 +44,49 @@
...
+				}
+				if(!filter.getCompetencyArea().isEmpty()){
+					if ((this.listToLowerCase(filter.getCompetencyArea()).contains(this.stringToLowerCase(eachInfo.getCompetencyArea())))){
+						afterFilter.add(eachInfo);
+					}
+				}
+			}
+		}
+		return afterFilter;
+	}
```
karthik-tarento: Looks like there is a possibility of having duplicate objects.
Filter contains - name:nameOne, type:typeOne, area:areaOne
Imagine there are 15 objects with 'nameOne'... those gets added into afterFilter. 
if there are 30 objects with "typeOne" -- which also has few of "nameOne" objects - both gets added into the list ??

src/main/java/org/sunbird/assessment/service/CohortsServiceImpl.java
```diff
@@ -337,6 +337,7 @@ private void processChildContentId(String givenContentId, List<String> assessmen
 					if (participantMap.containsKey(userId)) {
 						OpenSaberApiUserProfile userProfile = (OpenSaberApiUserProfile) participantMap.get(userId);
 						CohortUsers user = new CohortUsers();
+						user.setUserId(userProfile.getUserId());
```
karthik-tarento, line 339: This is not the proper solution... the API getting called is needs to be replaced with SB User Search API.

karthik-tarento (changes requested):

PR #33, opened 2021-11-29
Redis Level 2 Enhancement in sb-cb-ext

sreeragksgh:


src/main/resources/application.properties
```diff
@@ -127,3 +127,4 @@ azure.type.name=azure
 #Redis Cache Service properties
 redis.host.name=redisHostName
 redis.port=6379
+redis.timeout=redisTimeOutNeeded
```
karthik-tarento: Provide 84600 seconds (i.e. one day) as default value... 
add an comment above and mention that timeout value is in seconds...

src/main/java/org/sunbird/cache/RedisCache.java
```diff
@@ -3,22 +3,28 @@
...
 public class RedisCache {
 
     @Autowired
     private RedisTemplate redisTemplate;
 
+    @Autowired
+    CbExtServerProperties cbExtServerProperties;
+
     private CbExtLogger logger = new CbExtLogger(getClass().getName());
     private static final String KEY = "CB_EXT_";
```
karthik-tarento: We have Constants.REDIS_COMMON_KEY -- use the same instead of this..

src/main/java/org/sunbird/cache/service/RedisCacheServiceImpl.java
```diff
@@ -0,0 +1,90 @@
...
+import org.sunbird.common.model.SBApiResponse;
+import org.sunbird.common.util.Constants;
+import org.sunbird.core.logger.CbExtLogger;
+
+import java.util.*;
+
+public class RedisCacheServiceImpl implements RedisCacheService {
+
+    @Autowired
+    private RedisTemplate redisTemplate;
```
karthik-tarento: redisTemplate is not used - we can remove this.

karthik-tarento (changes requested):

karthik-tarento (approved):

PR #34, opened 2021-12-02
Updating the Budget apis for Datastax driver for Cassandra

FsherinP:


src/main/java/org/sunbird/budget/service/BudgetServiceImpl.java
```diff
@@ -44,51 +55,52 @@ public SBApiResponse submitBudgetDetails(BudgetInfo data, String userId) throws
...
+				request.put(Jsonkey.PROOF_DOCS, data.getProofdocs());
+				request.put(Jsonkey.SALARY_BUDGET_ALLOCATED, data.getSalaryBudgetAllocated());
+				request.put(Jsonkey.TRAINING_BUDGET_ALLOCATED, data.getTrainingBudgetAllocated());
+				request.put(Jsonkey.TRAINING_BUDGET_UTILIZATION, data.getTrainingBudgetUtilization());
+				
+				cassandraOperation.insertRecord(Jsonkey.DATABASE, Jsonkey.BUDGET_TABLE , request);	
+				cassandraOperation.insertRecord(Jsonkey.DATABASE, Jsonkey.AUDIT_TABLE, getAuditMap(userId, request, Constants.CREATE));
+				
+				response.getParams().setStatus(Constants.SUCCESSFUL);
+				response.put(Constants.DATA, request);
```
karthik-tarento, line 73: We should read the inserted record again and return it... not the request itself... db record will have proper Id values which is required for UI.

src/main/java/org/sunbird/cassandra/utils/CassandraPropertyReader.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.context.annotation.PropertySource;
+
+/**
+ * This class will be used to read cassandratablecolumn properties file.
+ * @author fathima
+ *
+ */
+
+@PropertySource("/src/main/resources:cassandratablecolumn.properties")
+public class CassandraPropertyReader {
```
karthik-tarento, line 16: Declare this as Configuration class. No need to add constructor or getInstance methods.
If you are using PropertySource - then you should be able to read it like env.getProperty(key)..

src/main/java/org/sunbird/common/util/Jsonkey.java
```diff
@@ -0,0 +1,31 @@
+package org.sunbird.common.util;
+
+/**
+ * This class contains all the key related to the cassandra table columns
+ * 
+ * @author fathima
+ *
+ */
+public final class Jsonkey {
```
karthik-tarento: This class is not required... just add the missing strings in Constants class.

src/main/resources/cassandratablecolumn.properties
```diff
@@ -0,0 +1,14 @@
+id=id
```
karthik-tarento, line 6: Add few lines of comment mentioning what is this file and how to use it.. 
It should mention purpose of the key and value.

karthik-tarento (changes requested):

PR #35, opened 2021-12-09
Staff repo cassandra driver update

Nivetha-M:


src/main/java/org/sunbird/staff/repo/StaffRepository.java
```diff
@@ -21,6 +21,6 @@
 	@Query(" delete from org_staff_position where orgId=?0 and id=?1")
     public List<StaffInfoModel> deleteStaffDetails(String rootOrg, String id);
 	
-	@Query("select * from org_staff_position where orgId=?0 and position=?1")
+	@Query("select * from org_staff_position where orgId=?0 and position=?1 ALLOW FILTERING")
```
karthik-tarento: We suppose to remove StaffRepository since we use CassandraOperation?
Also, never use "Allow Filtering" in cql query - it would cause performance issue..

src/main/java/org/sunbird/staff/model/StaffInfo.java
```diff
@@ -1,11 +1,15 @@
 package org.sunbird.staff.model;
 
+import com.fasterxml.jackson.annotation.JsonAlias;
+
 public class StaffInfo {
 
 	private String position;
 	private String orgId;
 	private String id;
+	@JsonAlias("totalpositionsfilled")
```
karthik-tarento: I guess these changes should go into cassandratablecolumn.properties

karthik-tarento (changes requested):

PR #44, opened 2022-01-06
user patch Profile API

sreeragksgh:


src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    RestTemplate restTemplate;
+
+    public static JSONObject transitionRequest = new JSONObject();
+    public static JSONArray transitionKeys = new JSONArray();
+    public static JSONObject profileDetails = new JSONObject();
+    public static JSONObject profile = new JSONObject();
+    public static JSONObject readData = new JSONObject();
+    public static JSONArray updatedField = new JSONArray();
+    public static JSONObject requestData = new JSONObject();
+    public static JSONObject transitionFrom = new JSONObject();
```
karthik-tarento, line 37: We should create objects only when necessary -- just before using them. Not in the class - unless its used by many methods.

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+
+@Component
+@Service
+public class ProfileServiceImpl implements ProfileService{
+
+    @Autowired
+    CbExtServerProperties serverConfig;
+
+    @Autowired
+    RestTemplate restTemplate;
```
karthik-tarento, line 28: Don't use restTemplate. use the OutboundRequestHandlerServiceImpl class -- that has more options and also logging.

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    private CbExtLogger logger = new CbExtLogger(getClass().getName());
+
+    @Override
+    public SBApiResponse profileUpdate(String request) throws Exception {
+
+        SBApiResponse response = new SBApiResponse(Constants.API_PROFILE_UPDATE);
+        try {
+            HttpHeaders headers = new HttpHeaders();
+            HttpEntity<Object> entity = new HttpEntity<Object>(headers);
+            String read = restTemplate.exchange(serverConfig.getCourseServiceHost()+serverConfig.getLmsUserReadPath(), HttpMethod.GET, entity, String.class).getBody();
```
karthik-tarento, line 47: Looks like this is wrong - you are not passing the userId to get the user details ?
Use Map<String, Object> to parse the return value.

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            readData = new JSONObject(read);
+            String approvalFields = restTemplate.exchange(serverConfig.getCourseServiceHost()+serverConfig.getLmsSystemSettingsPath(), HttpMethod.GET, entity, String.class).getBody();
+            requestData = new JSONObject(request);
+            JSONObject approvalSet = new JSONObject(approvalFields);
+            JSONObject approvalResult = (JSONObject) approvalSet.get(Constants.RESULT);
+            JSONObject approvalResponse = (JSONObject) approvalResult.get(Constants.RESPONSE);
+            String value = (String) approvalResponse.get(Constants.VALUE);
+            JSONArray profileConfig = new JSONArray(value);
```
karthik-tarento, line 55: This needs to be optimised more - move this to separate method to get the configuration. 
Use Map<String, Object> instead of JSONObject.

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            profile = (JSONObject) requestData.get("request");
+            profileDetails = (JSONObject) profile.get("profileDetails");
+            String userId = (String) profileDetails.get(Constants.USER_ID);
+            String deptName = (String) profileDetails.get(Constants.DEPT_NAME);
```
karthik-tarento, line 59: Looks like this is not correct.. 
Profile update request would contain userId inside request - not inside profileDetails ??
deptName won't be available in the profileDetails too ?

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+            String key = keys.next();
+            Object value = jsonObj.get(key);
+            if (value instanceof JSONArray) {
+                value = toList((JSONArray) value);
+            } else if (value instanceof JSONObject) {
+                value = toMap((JSONObject) value);
+            }
+            map.put(key, value);
+        }   return map;
+    }
```
karthik-tarento, line 192: Not required for this - if we simply use Map<String, Object> instead of json

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+    public static JSONObject profileDetails = new JSONObject();
+    public static JSONObject profile = new JSONObject();
+    public static JSONObject readData = new JSONObject();
+    public static JSONArray updatedField = new JSONArray();
+    public static JSONObject requestData = new JSONObject();
+    public static JSONObject transitionFrom = new JSONObject();
+    private CbExtLogger logger = new CbExtLogger(getClass().getName());
+
+    @Override
+    public SBApiResponse profileUpdate(String request) throws Exception {
```
karthik-tarento, line 41: This method needs more optimisation

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -0,0 +1,210 @@
...
+import org.sunbird.common.model.SBApiResponse;
+import org.sunbird.common.util.CbExtServerProperties;
+import org.sunbird.common.util.Constants;
+import org.sunbird.core.logger.CbExtLogger;
+
+import java.util.*;
+
+@Component
+@Service
+public class ProfileServiceImpl implements ProfileService{
```
karthik-tarento, line 22: Why this class is annotated with both Component and Service ?
Why are we are still using constant strings directly -- like "update", "transition" etc ? this suppose to be from Constants class? 
Need to be more careful whenever we create new object. Need to check how to avoid using temp variables and be productive -- otherwise it would be an performance issue.

karthik-tarento (changes requested):
Need more optimisation

PR #46, opened 2022-01-12
Cbrelease 3.0.2 rating review features

wilkysingh-fortum:
Add rating code is implemented.


src/main/java/org/sunbird/ratings/controller/RatingsController.java
```diff
@@ -0,0 +1,48 @@
...
+@RestController
+public class RatingsController {
+
+    @Autowired
+    RatingService ratingService;
+
+    // ----------------- Public APIs --------------------
+
+    @PostMapping("/ratings/v1/upsert")
+    public ResponseEntity<?> upsert(@Valid @RequestBody RequestRating requestRatingBody) throws Exception {
```
mathewjpallan: Should we just call this the update API?

src/main/java/org/sunbird/ratings/controller/RatingsController.java
```diff
@@ -0,0 +1,48 @@
...
+//    @GetMapping("/ratings/v1/read")
+//    public ResponseEntity<List<String>> getRatingsList() {
+//        return new ResponseEntity<>(ratingService.getRatings(activity_Id, activity_Type, userId), HttpStatus.OK);
+//    }
+//    @GetMapping("/ratings/v1/search")
+//    public ResponseEntity<List<String>> getUsersList() {
+//        return new ResponseEntity<>(ratingService.getUsers(activity_Id, activity_Type, userId), HttpStatus.OK);
+//    }
+
+//    @PostMapping("/ratings/v1/update")
```
mathewjpallan: This API is not required anymore

src/main/java/org/sunbird/ratings/exception/ValidateRatingException.java
```diff
@@ -0,0 +1,24 @@
+package org.sunbird.ratings.exception;
+
+
+import org.springframework.http.HttpStatus;
+import org.springframework.web.bind.annotation.ResponseBody;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+@ResponseStatus(value = HttpStatus.BAD_REQUEST)
+@ResponseBody
+public class ValidateRatingException extends RuntimeException {
```
mathewjpallan: Can we just have this class as ValidationException and introduce a list of error codes and have a list of messages also. Any API wishing to send a validation error should be able to reuse this. The reason for asking list of messages is because an API may want to perform all validations and give out a list of validation issues

src/main/java/org/sunbird/ratings/model/RatingMessage.java
```diff
@@ -0,0 +1,74 @@
+package org.sunbird.ratings.model;
+
+public class RatingMessage {
```
mathewjpallan, line 3: Can we not reuse one bean class instead of RatingMessage and RatingRequest? Both can have the same params right?

src/main/java/org/sunbird/ratings/service/RatingService.java
```diff
@@ -0,0 +1,17 @@
...
+public interface RatingService {
+    public SBApiResponse getRatings(String activity_Id, String activity_Type, String userId) throws Exception; //To do
+
+    public SBApiResponse getUsers(String activity_Id, String activity_Type, String userId) throws Exception;//To do
+
+    public SBApiResponse upsertRating(String activity_Id, String activity_Type, String userId) throws Exception; //To do
+
+    public SBApiResponse getRatingSummary(String activity_Id, String activity_Type) throws Exception; //To do
+
+    public SBApiResponse upsert(RequestRating requestRating) throws Exception;
```
mathewjpallan: We should never throw Exception which is a catchall. We need not even declare a throws clause as the custom exception thrown is a runtime exception.

src/main/java/org/sunbird/ratings/model/UpdatedValues.java
```diff
@@ -0,0 +1,32 @@
+package org.sunbird.ratings.model;
+
+public class UpdatedValues {
```
mathewjpallan: Is there a neat way to include this in the Rating bean class itself? Having a separate POJO does not seem like a good idea

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -0,0 +1,174 @@
...
+                response.setResponseCode(HttpStatus.OK);
+                return response;
+            }
+
+            request.put(Constants.CREATED_ON, timeBasedUuid);
+            request.put(Constants.RATING, requestRating.getRating());
+            request.put(Constants.REVIEW, requestRating.getReview());
+            request.put(Constants.UPDATED_ON, timeBasedUuid);
+
+            cassandraOperation.insertRecord(Constants.KEYSPACE_SUNBIRD, Constants.TABLE_RATINGS, request);
```
mathewjpallan: is the insertRecord function doing an upsert if the row is already present?

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -0,0 +1,174 @@
...
+
+            response.getParams().setStatus(Constants.SUCCESSFUL);
+            response.put(Constants.DATA, request);
+            response.setResponseCode(HttpStatus.CREATED);
+        } catch (ValidateRatingException ex) {
+            String errMsg = "Exception is : ";
+            return processExceptionBody(response, ex, errMsg, HttpStatus.BAD_REQUEST);
+        } catch (Exception ex) {
```
mathewjpallan, line 142: We should not have a catch block for Exception which is very broad.

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
mathewjpallan: or does the updateRecord add the record if it does not exist? in that case you do not need 2 sep calls to insert and update.

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -0,0 +1,174 @@
...
+                Map<String, Object> updateRequest = new HashMap<>();
+                updateRequest.put(Constants.RATING, requestRating.getRating());
+                updateRequest.put(Constants.REVIEW, requestRating.getReview());
+                updateRequest.put(Constants.UPDATED_ON, timeBasedUuid);
+
+                Map<String, Object> prevInfo = existingDataList.get(0);
+                cassandraOperation.updateRecord(Constants.KEYSPACE_SUNBIRD, Constants.TABLE_RATINGS, updateRequest,
+                        request);
+
+                msg.setAction("ratingUpdate");
```
mathewjpallan: We should have a constructor that takes in the basic types so that we can avoid so many setter calls

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -0,0 +1,174 @@
...
+public class RatingServiceImpl implements RatingService {
+    private ObjectMapper mapper = new ObjectMapper();
+
+    private Logger logger = LoggerFactory.getLogger(getClass().getName());
+
+    @Autowired
+    CassandraOperation cassandraOperation;
+
+    @Autowired
+    Producer producer;
```
mathewjpallan: Rename variable as kafka producer

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -0,0 +1,174 @@
...
+                UpdatedValues prevValues = processEventMessage(String.valueOf(prevInfo.get("createdon")),
+                        (Float) prevInfo.get("rating"), (String) prevInfo.get("review"));
+
+                UpdatedValues updatedValues = processEventMessage(String.valueOf(updateRequest.get(Constants.UPDATED_ON)),
+                        requestRating.getRating(), requestRating.getReview());
+
+                msg.setPrevValues(prevValues);
+                msg.setUpdatedValues(updatedValues);
+
+                producer.push(Constants.RATINGS_UPDATE_EVENT, msg);
```
mathewjpallan: The topic name should be read from env variables and not from constants. This is required because the topic names would be namespaced with env names like dev.topicname/stage.topicname etc

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -0,0 +1,174 @@
...
+                response.setResponseCode(HttpStatus.CREATED);
+                response.setResponseCode(HttpStatus.OK);
+                return response;
+            }
+
+            request.put(Constants.CREATED_ON, timeBasedUuid);
+            request.put(Constants.RATING, requestRating.getRating());
+            request.put(Constants.REVIEW, requestRating.getReview());
+            request.put(Constants.UPDATED_ON, timeBasedUuid);
+
```
mathewjpallan: The event needs to be sent to Kafka even if prevValues do not exist.

mathewjpallan (commented):
Please address these comments

src/main/java/org/sunbird/ratings/controller/RatingsController.java
wilkysingh-fortum: We are doing both addition and update in this so named it upsert to give more clarity about its working.

src/main/java/org/sunbird/ratings/controller/RatingsController.java
wilkysingh-fortum: Deleted.

src/main/java/org/sunbird/ratings/exception/ValidateRatingException.java
wilkysingh-fortum: These seems good add on.Will do it

src/main/java/org/sunbird/ratings/model/RatingMessage.java
wilkysingh-fortum: no there are couple of different params for example we are not maintaining version etc. and also have extra previous values and updated values attributes

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: ok

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: done

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: done

wilkysingh-fortum: Done

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: Done

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: No It doesn't validate whether the row already exists.

src/main/java/org/sunbird/ratings/model/UpdatedValues.java
wilkysingh-fortum: Done

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: done

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
wilkysingh-fortum: as discussed leaving it as it can catch other cassandra or unexpected sorts of errors

src/main/java/org/sunbird/ratings/service/RatingService.java
wilkysingh-fortum: done

mathewjpallan (approved):

PR #47, opened 2022-01-12
Added user course completion percentage in getUserProgress API

Nivetha-M:


src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -225,39 +245,54 @@ private UserProgressRequest validateGetBatchEnrolment(SunbirdApiRequest requestB
...
+
+	private int getCompletionStatus(int progress, int leafNodeCount) {
+		if (progress == 0) {
+			return 0;
+		} else if (progress >= 1 && progress < leafNodeCount) {
+			return 1;
+		} else {
+			return 2;
+		}
+	}
```
karthik-tarento: this code is duplicated. Just have one method which accepts the responseObj too... 
get the progress from responseObj and compare progress & leafNodeCount -- update the progress and status directly to responseObj

src/main/java/org/sunbird/common/model/SunbirdApiHierarchyResultContent.java
```diff
@@ -22,6 +22,8 @@
 	private String streamingUrl;
 	private List<SunbirdApiHierarchyResultContent> children;
 	private List<SunbirdApiBatchResp> batches;
+	private List<String> leafNodes;
```
karthik-tarento: We don't need to store the leafNodes details from hierarchyReponse object. 
leafNodes are child resource details for the particular course - which is not required.
Also - this not used anywhere else.

karthik-tarento (changes requested):

karthik-tarento (approved):

PR #57, opened 2022-03-15
Read Assessment API related changes

juhiagl8:


src/main/java/org/sunbird/assessment/model/Child.java
```diff
@@ -0,0 +1,267 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonPropertyOrder;
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonPropertyOrder({ "parent", "children", "name", "identifier", "description", "trackable", "primaryCategory",
+		"versionKey", "mimeType", "code", "version", "objectType", "status", "index", "maxQuestions" })
+
+public class Child extends JdkSerializationRedisSerializer implements Serializable{
```
karthik-tarento: Class name doesn't make any sense

src/main/java/org/sunbird/assessment/controller/AssessmentController.java
```diff
@@ -109,4 +110,18 @@
...
+	 * 
+	 * @param assessmentIdentifier
+	 * @param rootOrg
+	 * @return
+	 * @throws Exception
+	 */
+
+	@GetMapping("/v1/quml/assessment/read/{assessmentIdentifier}")
+	public ResponseEntity<Response> readAssessment(@PathVariable("assessmentIdentifier") String assessmentIdentifier,
+			@RequestHeader("rootOrg") String rootOrg, @RequestHeader("x-authenticated-user-token") String token) {
```
karthik-tarento: rootOrg is not required
assessmentIdentifier, x-authenticated-user-token - should be constant

src/main/java/org/sunbird/assessment/model/Children.java
```diff
@@ -0,0 +1,1099 @@
...
+    "contentEncoding",
+    "depth",
+    "allowAnonymousAccess",
+    "contentDisposition",
+    "visibility",
+    "showSolutions",
+    "index"
+})
+
+public class Children {
```
karthik-tarento: Class name doesn't make any sense

src/main/java/org/sunbird/assessment/model/QuestionSet1.java
```diff
@@ -0,0 +1,40 @@
+package org.sunbird.assessment.model;
+
+import java.io.Serializable;
+import java.util.List;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class QuestionSet1 implements Serializable {
```
karthik-tarento: Class name is not proper

src/main/java/org/sunbird/assessment/model/Response.java
```diff
@@ -0,0 +1,191 @@
...
+@JsonPropertyOrder({
+    "id",
+    "ver",
+    "ts",
+    "params",
+    "responseCode",
+    "result"
+})
+
+public class Response extends JdkSerializationRedisSerializer implements Serializable{
```
karthik-tarento: This is duplicate of src/main/java/org/sunbird/common/model/SBApiResponse.java

src/main/java/org/sunbird/assessment/service/AssessmentServiceImpl.java
```diff
@@ -272,4 +283,52 @@
...
+			}
+		} catch (Exception e) {
+			throw new ApplicationLogicError("REQUEST_COULD_NOT_BE_PROCESSED", e);
+		}
+		return new Response();
+	}
+
+	private Response extractMaxQuestionsAndCreateResultSet(String assessmentIdentifier, Map<String, Object> result,
+			Response response) {
+		if (!ObjectUtils.isEmpty(response.getResult().getQuestionSet().getChildren())) {
```
karthik-tarento: This check is unnecessary - this method should be called only when the result and questionset children is not empty.

src/main/java/org/sunbird/assessment/service/AssessmentServiceImpl.java
```diff
@@ -272,4 +283,52 @@
 		return result;
 	}
 
+	@Override
+	public Response readAssessment(String assessmentIdentifier, String token) {
```
karthik-tarento: Code logic needs to be improved in this method.
resp -- should not be empty when returning... this method should either return successful assessment details or should throw 404, 500 errors.

src/main/java/org/sunbird/assessment/service/AssessmentServiceImpl.java
```diff
@@ -272,4 +283,52 @@
...
+	}
+
+	private Response extractMaxQuestionsAndCreateResultSet(String assessmentIdentifier, Map<String, Object> result,
+			Response response) {
+		if (!ObjectUtils.isEmpty(response.getResult().getQuestionSet().getChildren())) {
+			Iterator<Child> sections = response.getResult().getQuestionSet().getChildren().iterator();
+			while (sections.hasNext()) {
+				{
+					Child section = sections.next();
+					if (section.getParent().equals(Constants.DO_QNS_SET + assessmentIdentifier)) {
```
karthik-tarento: this if condition is unnecessary...
when adding an if condition code must have else part too... in this case, if child's parent id is not same as assessment id - then it should be treated an issue and either through error or log the error
No need to prefix with "do_" -- identifier always start with do_

src/main/java/org/sunbird/assessment/service/CohortsServiceImpl.java
```diff
@@ -18,7 +18,15 @@
 import org.springframework.util.StringUtils;
 import org.sunbird.assessment.repo.CohortUsers;
 import org.sunbird.assessment.repo.UserAssessmentTopPerformerRepository;
-import org.sunbird.common.model.*;
+import org.sunbird.common.model.OpenSaberApiUserProfile;
```
karthik-tarento: OpenSaber related classes should not be used.

karthik-tarento (changes requested):
Instead of models -- we need to explore the options to read the parameters based on configuration.

PR #58, opened 2022-03-16
Added the user's last login API for audit telemetry events

juhiagl8:


karthik-tarento (changes requested):
1. DeptPublicInfo -- need to be in portal.department.model package only.
2. cb.ext.version should be 4.0.1
3. Use _ in constant variable names when starting new word

PR #64, opened 2022-04-08
lastname_firstname added and refactoring

wilkysingh-tarento:


src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -76,28 +74,89 @@ public SBApiResponse getRatings(String activity_Id, String activity_Type, String
...
-            request.put(Constants.ACTIVITY_ID, activity_Id);
-            request.put(Constants.ACTIVITY_TYPE, activity_Type);
+            request.put(Constants.ACTIVITY_ID, activityId);
+            request.put(Constants.ACTIVITY_TYPE, activityType);
 
             List<Map<String, Object>> existingDataList = cassandraOperation.getRecordsByProperties(Constants.KEYSPACE_SUNBIRD,
                     Constants.TABLE_RATINGS_SUMMARY, request, null);
+
             if (!CollectionUtils.isEmpty(existingDataList)) {
+            Map<String, Object> map = existingDataList.get(0);
```
karthik-tarento: The same is used again below -- declare this above and avoid creating new variable again

src/main/java/org/sunbird/ratings/service/RatingServiceImpl.java
```diff
@@ -76,28 +74,89 @@ public SBApiResponse getRatings(String activity_Id, String activity_Type, String
...
+                if (userModel.getLastName() != null) {
+                    lastName = userModel.getLastName();
+                }
+                latest50Reviews.add(new SummaryModel.latestReviews(Constants.REVIEW,
+                        userModel.getId(),
+                        summaryNodeModel.getDate(),
+                        summaryNodeModel.getRating().floatValue(),
+                        summaryNodeModel.getReview(),
+                        firstName,
+                        lastName
```
karthik-tarento: variable will hold the previous iteration values if null... avoid variable and directly pass the value by checking condition

karthik-tarento (changes requested):

PR #114, opened 2022-08-08
Assessment final

juhiagl8:


src/main/java/org/sunbird/assessment/repo/AssessmentRepositoryImpl.java
```diff
@@ -138,13 +138,13 @@ public boolean addUserAssesmentStartTime(String userId, String assessmentIdentif
...
+	public Date fetchUserAssessmentStartTime(String userId, String assessmentIdentifier) {
 		Map<String, Object> request = new HashMap<>();
 		request.put(Constants.USER_ID, userId);
 		request.put(Constants.IDENTIFIER, assessmentIdentifier);
-		Map<String, Object> existingDataList = cassandraOperation.getRecordsByProperties(Constants.KEYSPACE_SUNBIRD,
-				Constants.TABLE_USER_ASSESSMENT_TIME, request, null).get(0);
-		return (Date) existingDataList.get("starttime");
+		List<Map<String, Object>> existingDataList = cassandraOperation.getRecordsByProperties(Constants.KEYSPACE_SUNBIRD,
+				Constants.TABLE_USER_ASSESSMENT_TIME, request, null);
+		return (!existingDataList.isEmpty()) ? (Date) existingDataList.get(0).get("starttime") : null;
```
karthik-tarento: So - we will have more than one data for given userId and assessmentId ?
getting the 1st value from list -- is correct? is the cassandra query will return sorted list ?

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2.java
```diff
@@ -14,7 +14,9 @@ public interface AssessmentServiceV2 {
 	 */
 	public SBApiResponse submitAssessment(Map<String, Object> data, String userEmail) throws Exception;
 
-	public SBApiResponse readAssessment(String assessmentIdentifier, String token) throws Exception;
+	public SBApiResponse readAssessment(String assessmentIdentifier, String token);
 
-	public SBApiResponse readQuestionList(Map<String, Object> requestBody, String authUserToken) throws Exception;
+	public SBApiResponse readQuestionList(Map<String, Object> requestBody, String authUserToken);
+
+	Map<String, Object> createResponseMapWithProperStructure(Map<String, Object> hierarchySection, Map<String, Object> resultMap);
```
karthik-tarento: why this method needs to be added in interface ??

src/main/resources/application.properties
```diff
@@ -68,6 +67,7 @@ kafka.topics.parent.telemetry.event=dev.telemetry.raw
 kafka.topics.parent.rating.event = dev.rating.event
 kafka.topics.userutility.telemetry.event=userLastLoginTopic
 kafka.topics.wat.telemetry.event.topic.consumer=watTelemetryTopic-consumer
+kafka.topics.user.assessment.submit=local.dev.assessment.submit
```
karthik-tarento: suggestion - avoid "local" or "dev" in the topic name -- it doesn't make sense.. just -- "assessment.submit" or "assessment.result.submit"

src/main/resources/application.properties
```diff
@@ -140,13 +140,13 @@ redis.port=6379
...
+#assessment.host=http://assessment-service:9000/
+#assessment.hierarchy.read.path=questionset/v4/hierarchy/{identifier}?mode=edit
+#assessment.question.list.path=question/v4/list
 
-#assessment.host=https://igot-dev.in/api/
-#assessment.hierarchy.read.path=questionset/v1/hierarchy/{identifier}?hierarchy=detail
-#assessment.question.list.path=question/v/list
+assessment.host=https://igot-dev.in/api/
+assessment.hierarchy.read.path=questionset/v1/hierarchy/{identifier}?mode=edit
+assessment.question.list.path=question/v1/list
```
karthik-tarento: changes done for local - should not be committed...

src/main/java/org/sunbird/scheduler/util/EmailNotificationService.java
```diff
@@ -30,11 +30,11 @@ public static void sendIncompleteCourseEmail(Map.Entry<String, UserCourseProgres
...
-                    params.put(Constants.COURSE + j + Constants._NAME, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCourseName());
-                    params.put(Constants.COURSE + j + Constants._DURATION, String.valueOf(userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCompletionPercentage()));
+                    params.put(Constants.COURSE_KEY + j, true);
+                    params.put(Constants.COURSE_KEY + j + Constants._URL, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCourseUrl());
+                    params.put(Constants.COURSE_KEY + j + Constants.THUMBNAIL, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getThumbnail());
+                    params.put(Constants.COURSE_KEY + j + Constants._NAME, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCourseName());
+                    params.put(Constants.COURSE_KEY + j + Constants._DURATION, String.valueOf(userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCompletionPercentage()));
```
karthik-tarento: can you raise this as separate PR - so this can be reviewed and released much faster

src/main/java/org/sunbird/common/util/Constants.java
```diff
@@ -335,6 +338,7 @@ public class Constants {
 	public static final String SUNBIRD_CASSANDRA_CONSISTENCY_LEVEL = "sunbird_cassandra_consistency_level";
 	public static final String INCOMPLETE_COURSES_MAIL_SUBJECT = "Complete the courses you started";
 	public static final String COURSE = "Course";
+	public static final String COURSE_KEY = "course";
```
karthik-tarento: In these cases -- when you require a constant with all small cases... 
just check the word exist -- if so, simply use that like --> COURSE.toLowerCase() -- instead of creating one more constant ?

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
     @Autowired
     RedisCacheMgr redisCacheMgr;
 
+    @Autowired
+    CbExtServerProperties serverProperties;
+
     @Autowired
     CbExtServerProperties cbExtServerProperties;
```
karthik-tarento: twice ??

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
             if (userId != null) {
                 Map<String, Object> assessmentAllDetail = (Map<String, Object>) redisCacheMgr
                         .getCache(Constants.ASSESSMENT_ID + assessmentIdentifier);
-                boolean isSuccess = true;
                 if (ObjectUtils.isEmpty(assessmentAllDetail)) {
                     Map<String, Object> hierarcyReadApiResponse = getReadHierarchyApiResponse(assessmentIdentifier, token);
-                    if (!Constants.OK.equalsIgnoreCase((String) hierarcyReadApiResponse.get(Constants.RESPONSE_CODE))) {
-                        isSuccess = false;
+                    if (hierarcyReadApiResponse.isEmpty() || !Constants.OK.equalsIgnoreCase((String) hierarcyReadApiResponse.get(Constants.RESPONSE_CODE))) {
+                        errMsg = "Please check the Read Hierarchy API";
```
karthik-tarento: Need to update errMsg --> proper one should be like... 
"Assessment hierarchy read failed, failed to process Assessment Read Request"

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
                                 .get(Constants.RESULT)).get(Constants.QUESTION_SET);
                         redisCacheMgr.putCache(Constants.ASSESSMENT_ID + assessmentIdentifier, assessmentAllDetail);
                     }
                 }
-                response = prepareAssessmentResponse(assessmentAllDetail, isSuccess);
-                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + token, response.getResult().get(Constants.QUESTION_SET));
-                if (assessmentAllDetail.get(Constants.DURATION) != null) {
-                    boolean resp = assessmentRepository.addUserAssesmentStartTime(userId, Constants.ASSESSMENT_ID + assessmentIdentifier, new Timestamp(new Date().getTime()));
-                    return response;
+                response.getResult().put(Constants.QUESTION_SET, readAssessmentLevelData(assessmentAllDetail));
```
karthik-tarento: when Assessment Read called first time... 
assessmentAllDetail would be empty... in this case if we failed to call Assessment hierarchy -- what would happen here? looks like we are trying to store empty value into redisCache... check errMsg is not empty and then proceed...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + assessmentIdentifier + token, response.getResult().get(Constants.QUESTION_SET));
+                if (assessmentAllDetail.get(Constants.EXPECTED_DURATION) != null) {
```
karthik-tarento: What is the necessary for this if condition ?
do not assume courses without expected duration -- if so, then we need to through error -- should not ignore adding startTime

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
         } catch (Exception e) {
-            logger.error(e);
-            throw new ApplicationLogicError("REQUEST_COULD_NOT_BE_PROCESSED", e);
+            logger.error(String.format("Exception in %s : %s", "read Assessment", e.getMessage()), e);
+            errMsg = "Failed to read Assessment. Exception: " + e.getMessage();
+        }
+        if (StringUtils.isNotBlank(errMsg)) {
+            response.getParams().setStatus(Constants.FAILED);
+            response.getParams().setErrmsg(errMsg);
+            response.setResponseCode(HttpStatus.BAD_REQUEST);
```
karthik-tarento: HttpStatus should be InternalServerError -- Bad_Request can be used if it is failed when validating the request

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
                         redisCacheMgr.putCache(Constants.ASSESSMENT_ID + assessmentIdentifier, assessmentAllDetail);
                     }
                 }
-                response = prepareAssessmentResponse(assessmentAllDetail, isSuccess);
-                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + token, response.getResult().get(Constants.QUESTION_SET));
-                if (assessmentAllDetail.get(Constants.DURATION) != null) {
-                    boolean resp = assessmentRepository.addUserAssesmentStartTime(userId, Constants.ASSESSMENT_ID + assessmentIdentifier, new Timestamp(new Date().getTime()));
-                    return response;
+                response.getResult().put(Constants.QUESTION_SET, readAssessmentLevelData(assessmentAllDetail));
+                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + assessmentIdentifier + token, response.getResult().get(Constants.QUESTION_SET));
```
karthik-tarento: One more case - Need to check putCache is successful or not.. if failed - need to send back error response to user.

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+    private SBApiResponse createDefaultResponse(String api) {
+        SBApiResponse response = new SBApiResponse();
+        response.setId(api);
+        response.setVer(Constants.VER);
+        response.getParams().setResmsgid(UUID.randomUUID().toString());
+        response.getParams().setStatus(Constants.SUCCESS);
+        response.setResponseCode(HttpStatus.OK);
+        response.setTs(DateTime.now().toString());
+        return response;
+    }
```
karthik-tarento: suggestion - add new private methods at the end of the file... so that it is easy to review..
public methods (which are exposed to interface) need to be added at the end of previous public methods..

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
-                            redisCacheMgr.putCache(Constants.QUESTION_ID + qmap.get(Constants.IDENTIFIER), qmap);
-                            questionList.add(filterQuestionMapDetail(qmap));
-                        } else {
-                            logger.error(new Exception("Failed to get Question Details for Id: " + qmap.get(Constants.IDENTIFIER)));
+            String userId = RequestInterceptor.fetchUserIdFromAccessToken(authUserToken);
+            if (userId != null) {
+                List<String> identifierList = getQuestionIdList(requestBody);
+                List<Object> questionList = new ArrayList<>();
+                List<String> newIdentifierList = new ArrayList<>();
+                if (requestBody.containsKey(Constants.ASSESSMENT_ID_KEY) && !requestBody.get(Constants.ASSESSMENT_ID_KEY).toString().isEmpty() && !identifierList.isEmpty()) {
```
karthik-tarento: !requestBody.get(Constants.ASSESSMENT_ID_KEY).toString().isEmpty() -- bad... it should be... 
!StringUtils.isEmpty((String)requestBody.get(Constants.ASSESSMENT_ID_KEY))

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                    String key = Constants.USER_ASSESS_REQ + requestBody.get(Constants.ASSESSMENT_ID_KEY).toString() + authUserToken;
+                    Map<String, Object> questionSetFromAssessment = (Map<String, Object>) redisCacheMgr.getCache(key);
+                    if (questionSetFromAssessment != null && questionSetFromAssessment.get(Constants.CHILDREN) != null) {
```
karthik-tarento: We are storing the questionList into the cache... and we are 100% sure that an empty children is stored in Cache... 
so - just !ObjectUtils.isEmpty(questionSetFromAssessment) would be enough...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                        List<String> questionsFromAssessment = new ArrayList<>();
+                        List<Map<String, Object>> sections = (List<Map<String, Object>>) questionSetFromAssessment.get(Constants.CHILDREN);
+                        for (Map<String, Object> section : sections) {
+                            questionsFromAssessment.addAll((List<String>) section.get(Constants.CHILD_NODES));
+                        }
+                        //Out of the list of questions received in the payload, retaining only those ids which are a part of the user's latest assessment
+                        identifierList.retainAll(questionsFromAssessment);
```
karthik-tarento: if user send readQuestionList request with more questionId than what we actually provided in readAssessment response -- are we simply removing the extra id ?
we suppose to throw error -- right ?

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                        //Fetching all the remaining questions details from the Redis
+                        List<Object> map = redisCacheMgr.mget(identifierList);
+                        for (int i = 0; i < map.size(); i++) {
+                            if (ObjectUtils.isEmpty(map.get(i))) {
+                                //Adding the not found keys as a seperate list
+                                newIdentifierList.add(identifierList.get(i));
+                            } else {
+                                //Filtering the details of the questions which are found in redis and adding them to another list
+                                questionList.add(filterQuestionMapDetail((Map<String, Object>) map.get(i)));
+                            }
```
karthik-tarento: Are you sure that redisCacheMgr would return empty object if the given key is not found ??
say... identifierList is --> [1,2,3,4,5]
if 2 and 3 are not there, then redisCache would return -> [obj_1, null, null, obj_4, obj_5] -- ?? I guess it would return only [obj_1, obj_4, obj_5] ??

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                        }
+                        //Taking the list which was formed with the not found values in Redis, we are making an internal POST call to Question List API to fetch the details
+                        if (!newIdentifierList.isEmpty()) {
+                            Map<String, Object> questionMapResponse = readQuestionDetails(newIdentifierList);
+                            if (questionMapResponse != null && !questionMapResponse.isEmpty() && Constants.OK.equalsIgnoreCase((String) questionMapResponse.get(Constants.RESPONSE_CODE))) {
```
karthik-tarento: questionMapResponse != null && !questionMapResponse.isEmpty() --> !ObjectUtils.isEmpty(questionMapResponse)

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                                    if (!ObjectUtils.isEmpty(questionMap)) {
+                                        redisCacheMgr.putCache(Constants.QUESTION_ID + question.get(Constants.IDENTIFIER), question);
+                                        questionList.add(filterQuestionMapDetail(question));
+                                    } else {
+                                        logger.error(String.format("Failed to get Question Details for Id: %s", question.get(Constants.IDENTIFIER).toString()));
+                                    }
+                                }
+                            } else {
+                                logger.error(String.format("Failed to get Question Details from the Question List API for the IDs: %s", newIdentifierList.toString()));
+                            }
```
karthik-tarento: We just log error ? what if this is the 1st time, there won't be any information in the cache... in that case, we simply log error and return empty successful response ?
we should through error even if we are not return one question details

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -34,146 +39,329 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                        else
+                            errMsg = "The list of questions are Empty";
+                    } else {
+                        errMsg = "Please provide a valid assessment Id/Session Expired";
                     }
+                } else {
+                    errMsg = "Please provide the valid Assessment Id";
                 }
+            } else {
+                errMsg = "User Id doesn't exist! Please supply a valid auth token";
```
karthik-tarento: can we have a validation method for this readQuestionList... the validation method should check the following... 
a) check user token is valid
b) incoming request contains assessmentId and list of questionId's
c) check the cache for readAssessmentResponse - make sure the incoming request for readQuestionList contains Id's from what we sent in readAssessmentResponse
return error response if one of the above failed...

karthik-tarento (changes requested):

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
juhiagl8: yes, It returns null karthik, to preserve the order of the elements, otherwise anyone would get confused, so element 5 will have value in 5 as the response

src/main/resources/application.properties
```diff
@@ -145,8 +145,8 @@ assessment.hierarchy.read.path=questionset/v4/hierarchy/{identifier}?mode=edit
 assessment.question.list.path=question/v4/list
 
 #assessment.host=https://igot-dev.in/api/
-#assessment.hierarchy.read.path=questionset/v1/hierarchy/{identifier}?hierarchy=detail
-#assessment.question.list.path=question/v/list
+#assessment.hierarchy.read.path=questionset/v1/hierarchy/{identifier}?mode=edit
```
karthik-tarento: param should be --> hierarchyType=detail

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,348 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
                 }
-                response = prepareAssessmentResponse(assessmentAllDetail, isSuccess);
-                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + token, response.getResult().get(Constants.QUESTION_SET));
-                if (assessmentAllDetail.get(Constants.DURATION) != null) {
-                    boolean resp = assessmentRepository.addUserAssesmentStartTime(userId, Constants.ASSESSMENT_ID + assessmentIdentifier, new Timestamp(new Date().getTime()));
-                    return response;
+                response.getResult().put(Constants.QUESTION_SET, readAssessmentLevelData(assessmentAllDetail));
+                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + assessmentIdentifier + token, response.getResult().get(Constants.QUESTION_SET));
+                if (assessmentAllDetail.get(Constants.EXPECTED_DURATION) != null) {
+                    assessmentRepository.addUserAssesmentStartTime(userId, Constants.ASSESSMENT_ID + assessmentIdentifier, new Timestamp(new Date().getTime()));
```
karthik-tarento: Check DB write is successful - if failed return error response to user.

src/main/java/org/sunbird/scheduler/util/EmailNotificationService.java
```diff
@@ -30,11 +30,11 @@ public static void sendIncompleteCourseEmail(Map.Entry<String, UserCourseProgres
...
-                    params.put(Constants.COURSE + j + Constants._NAME, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCourseName());
-                    params.put(Constants.COURSE + j + Constants._DURATION, String.valueOf(userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCompletionPercentage()));
+                    params.put(Constants.COURSE.toLowerCase() + j, true);
+                    params.put(Constants.COURSE.toLowerCase() + j + Constants._URL, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCourseUrl());
+                    params.put(Constants.COURSE.toLowerCase() + j + Constants.THUMBNAIL, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getThumbnail());
+                    params.put(Constants.COURSE.toLowerCase() + j + Constants._NAME, userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCourseName());
+                    params.put(Constants.COURSE.toLowerCase() + j + Constants._DURATION, String.valueOf(userCourseProgressDetailsEntry.getValue().getIncompleteCourses().get(i).getCompletionPercentage()));
```
karthik-tarento: Need to be removed - as this is done in separate PR

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
                         redisCacheMgr.putCache(Constants.ASSESSMENT_ID + assessmentIdentifier, assessmentAllDetail);
                     }
                 }
-                response = prepareAssessmentResponse(assessmentAllDetail, isSuccess);
-                redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + token, response.getResult().get(Constants.QUESTION_SET));
-                if (assessmentAllDetail.get(Constants.DURATION) != null) {
-                    boolean resp = assessmentRepository.addUserAssesmentStartTime(userId, Constants.ASSESSMENT_ID + assessmentIdentifier, new Timestamp(new Date().getTime()));
-                    return response;
+                if(errMsg.isEmpty()) {
+                    response.getResult().put(Constants.QUESTION_SET, readAssessmentLevelData(assessmentAllDetail));
```
karthik-tarento, line 89: response object need to be updated after redis putCache is successful... 
otherwise we would send error response --- but still there would be assessment details in the response...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                    Boolean isInsertedToRedis = redisCacheMgr.putCache(Constants.USER_ASSESS_REQ + assessmentIdentifier + token, response.getResult().get(Constants.QUESTION_SET));
+                    if (!isInsertedToRedis) {
+                        errMsg = "Data couldn't be inserted to Redis, Please check!";
+                    }
+                    Boolean isStartTimeUpdated = assessmentRepository.addUserAssesmentStartTime(userId, Constants.ASSESSMENT_ID + assessmentIdentifier, new Timestamp(new Date().getTime()));
+                    if (!isStartTimeUpdated) {
+                        errMsg = "Assessment Start Time not updated! Please check!";
+                    }
```
karthik-tarento: if putCache is successful then only we need to try to add start time and assessment details in response... need else block for this... 
current code will try to add start time irrespective of successful or failure of putCache... which is wrong

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -311,89 +489,58 @@ private Map<String, Object> filterQuestionMapDetail(Map<String, Object> question
...
+        response.getParams().setStatus(Constants.SUCCESS);
+        response.setResponseCode(HttpStatus.OK);
+        response.setTs(DateTime.now().toString());
+        return response;
     }
 
+    private Boolean validateQuestionListRequest(List<String> identifierList, List<String> questionsFromAssessment) {
+        List<String> identifierListCopy = identifierList;
+        identifierListCopy.removeAll(questionsFromAssessment);
+        return identifierListCopy.isEmpty() ? Boolean.TRUE : Boolean.FALSE;
```
karthik-tarento: the intention of having separate method is to perform the basic functionalities here so that code looks readable... 
readQuestionList API again does the major work of checking the incoming request has AssessmentId and list of questions object... 
move those to here -- make sure all validation done here...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
-                Map<String, Object> questionMapResponse = readQuestionDetails(newIdentifierList);
-                if (questionMapResponse != null && Constants.OK.equalsIgnoreCase((String) questionMapResponse.get(Constants.RESPONSE_CODE))) {
-                    List<Map<String, Object>> questionMap = ((List<Map<String, Object>>) ((Map<String, Object>) questionMapResponse
-                            .get(Constants.RESULT)).get(Constants.QUESTIONS));
-                    for (Map<String, Object> qmap : questionMap) {
-                        if (!ObjectUtils.isEmpty(questionMap)) {
-                            redisCacheMgr.putCache(Constants.QUESTION_ID + qmap.get(Constants.IDENTIFIER), qmap);
-                            questionList.add(filterQuestionMapDetail(qmap));
+            String userId = validateAuthTokenAndFetchUserId(authUserToken);
+            if (userId != null && requestBody.containsKey(Constants.ASSESSMENT_ID_KEY) && !StringUtils.isEmpty((String)requestBody.get(Constants.ASSESSMENT_ID_KEY))) {
```
karthik-tarento: why do we have separate validation method and still this method checks requestBody contains this or that ?
move all the conditions to validation method... 
make the code more readable... 
example... 
if(StringUtils.isEmpty(userId) {
errMsg = "some error";
logger.error(errMsg);
//update response with status and errMsg
return response; 
}

current code has huge if block -- it's difficult to read or track where it starts and where it ends...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                List<String> identifierList = getQuestionIdList(requestBody);
+                if (!identifierList.isEmpty()) {
```
karthik-tarento: if these are verified in validation method and if any error returned as soon as we call validation method -- then we don't need to have this if block at all...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                    Map<String, Object> questionSetFromAssessment = (Map<String, Object>) redisCacheMgr.getCache(key);
+                    if (!ObjectUtils.isEmpty(questionSetFromAssessment)) {
+                        List<String> questionsFromAssessment = new ArrayList<>();
+                        List<Map<String, Object>> sections = (List<Map<String, Object>>) questionSetFromAssessment.get(Constants.CHILDREN);
+                        for (Map<String, Object> section : sections) {
+                            questionsFromAssessment.addAll((List<String>) section.get(Constants.CHILD_NODES));
+                        }
+                        //Out of the list of questions received in the payload, checking if the request has only those ids which are a part of the user's latest assessment
+                        //Fetching all the remaining questions details from the Redis
+                        if (validateQuestionListRequest(identifierList, questionsFromAssessment)) {
```
karthik-tarento: move these conditions into validation method

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+                                            Boolean isInsertedToRedis = redisCacheMgr.putCache(Constants.QUESTION_ID + question.get(Constants.IDENTIFIER), question);
+                                            questionList.add(filterQuestionMapDetail(question));
+                                        } else {
+                                            errMsg = "Failed to get Question Details for Id: %s";
+                                            logger.error(String.format("Failed to get Question Details for Id: %s", question.get(Constants.IDENTIFIER).toString()));
+                                            break;
+                                        }
+                                    }
+                                } else {
+                                    logger.error(String.format("Failed to get Question Details from the Question List API for the IDs: %s", newIdentifierList.toString()));
```
karthik-tarento: update errMsg in this else part too...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,364 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+            response.getParams().setErrmsg(errMsg);
+            response.setResponseCode(HttpStatus.BAD_REQUEST);
         }
+        return response;
+
+    }
 
+    private static String validateAuthTokenAndFetchUserId(String authUserToken) {
+        return RequestInterceptor.fetchUserIdFromAccessToken(authUserToken);
     }
```
karthik-tarento: Separate method is not required for single lines -- previous comment w.r.t. validation method is for validating incoming request object.
you can move this line also to the other validation method... check the userId is not empty in that method and return errMsg...

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -311,89 +489,58 @@ private Map<String, Object> filterQuestionMapDetail(Map<String, Object> question
...
+        response.setId(api);
+        response.setVer(Constants.VER);
+        response.getParams().setResmsgid(UUID.randomUUID().toString());
+        response.getParams().setStatus(Constants.SUCCESS);
+        response.setResponseCode(HttpStatus.OK);
+        response.setTs(DateTime.now().toString());
+        return response;
     }
 
+    private Boolean validateQuestionListRequest(List<String> identifierList, List<String> questionsFromAssessment) {
```
karthik-tarento, line 610: This method should return String (i.e. errMsg) -- in readQuestionList just check errMsg is not empty -- if so, return error response...

karthik-tarento (changes requested):

src/main/java/org/sunbird/core/config/RedisConfig.java
```diff
@@ -39,4 +39,5 @@ public RedisTemplate<String, Object> redisTemplate() {
 		redisTemplate.afterPropertiesSet();
 		return redisTemplate;
 	}
+
```
karthik-tarento, line 42: Remove this change

src/main/resources/application.properties
```diff
@@ -39,7 +39,6 @@ lms.user.update.path=/private/user/v1/update
 progress.api.endpoint=v1/content/state/read
 participants.api.endpoint=v1/batch/participants/list
 sb.api.key=apiKey
-
```
karthik-tarento, line 42: Remove this change

src/main/resources/application.properties
```diff
@@ -141,12 +141,12 @@ redis.timeout=84600
...
 assessment.host=http://assessment-service:9000/
-assessment.hierarchy.read.path=questionset/v4/hierarchy/{identifier}?mode=edit
+assessment.hierarchy.read.path=questionset/v4/hierarchy/{identifier}?hierarchyType=detail
 assessment.question.list.path=question/v4/list
 
 #assessment.host=https://igot-dev.in/api/
-#assessment.hierarchy.read.path=questionset/v1/hierarchy/{identifier}?hierarchy=detail
-#assessment.question.list.path=question/v/list
+#assessment.hierarchy.read.path=questionset/v1/hierarchy/{identifier}?hierarchyType=detail
+#assessment.question.list.path=question/v1/list
```
karthik-tarento: Remove the commented out properties... only explanation of property can be method in commented lines...

src/main/resources/application.properties
```diff
@@ -68,6 +67,7 @@ kafka.topics.parent.telemetry.event=dev.telemetry.raw
 kafka.topics.parent.rating.event = dev.rating.event
 kafka.topics.userutility.telemetry.event=userLastLoginTopic
 kafka.topics.wat.telemetry.event.topic.consumer=watTelemetryTopic-consumer
+kafka.topics.user.assessment.submit=assessment.submit
```
karthik-tarento, line 68: Make sure README file has information about this topic

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,372 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+        }
+        return response;
+
+    }
 
+    private String validateAuthTokenAndFetchUserId(String authUserToken) {
+        return RequestInterceptor.fetchUserIdFromAccessToken(authUserToken);
+    }
+
+    private String validateQuestionListAPI(Map<String, Object> requestBody, String authUserToken, String errMsg, List<String> identifierList) {
```
karthik-tarento: why errMsg passed here ?

src/main/java/org/sunbird/assessment/service/AssessmentServiceV2Impl.java
```diff
@@ -35,195 +40,372 @@ public class AssessmentServiceV2Impl implements AssessmentServiceV2 {
...
+        String userId = validateAuthTokenAndFetchUserId(authUserToken);
+        if (userId != null && requestBody.containsKey(Constants.ASSESSMENT_ID_KEY) && !StringUtils.isEmpty((String)requestBody.get(Constants.ASSESSMENT_ID_KEY))) {
```
karthik-tarento: This is bad...
if userId is not null and requestBody doesn't contain AssessmentId - still the error message would be -- ""User Id doesn't exist! Please supply a valid auth token" -- why ??

karthik-tarento (changes requested):

PR #165, opened 2023-01-05
Org create API enhancement to accept user entries of state/ministry/org/mdo.

sreeragksgh:


src/main/java/org/sunbird/org/service/ExtendedOrgServiceImpl.java
```diff
@@ -58,33 +58,65 @@ public SBApiResponse createOrg(Map<String, Object> request, String userToken) {
...
+						mapId = createMapId(requestData);
+						updateRequest.put(Constants.MAP_ID, mapId);
 					}
-				}
-				Map<String, Object> compositeKey = new HashMap<String, Object>() {
-					private static final long serialVersionUID = 1L;
-					{
-						put(Constants.ORG_NAME, requestData.get(Constants.ORG_NAME));
-						put(Constants.MAP_ID, requestData.get(Constants.MAP_ID));
+					if(requestData.containsKey(Constants.SB_ROOT_ORG_ID)){
```
karthik-tarento: should be checked against null -- instead of containsKey

src/main/java/org/sunbird/org/service/ExtendedOrgServiceImpl.java
```diff
@@ -58,33 +58,65 @@ public SBApiResponse createOrg(Map<String, Object> request, String userToken) {
...
-						updateRequest.put(Constants.SB_ROOT_ORG_ID, sbRootOrgId);
+				Map<String, Object> updateRequest = new HashMap<String, Object>();
+				updateRequest.put(Constants.SB_ORG_ID, orgId);
+				updateRequest.put(Constants.ORG_NAME, orgName);
+				updateRequest.put(Constants.SB_ORG_TYPE, orgType);
+				String mapId = null;
+				updateRequest.put(Constants.SB_SUB_ORG_TYPE, requestData.get(Constants.ORGANIZATION_SUB_TYPE));
+				if (!Constants.STATE.equalsIgnoreCase(orgType) && !Constants.MINISTRY.equalsIgnoreCase(orgType)) {
+					updateRequest.put(Constants.PARENT_MAP_ID, requestData.get(Constants.PARENT_MAP_ID));
+					if(requestData.containsKey(Constants.MAP_ID)){
```
karthik-tarento: should be checked against null -- instead of containsKey

src/main/java/org/sunbird/org/service/ExtendedOrgServiceImpl.java
```diff
@@ -58,33 +58,55 @@ public SBApiResponse createOrg(Map<String, Object> request, String userToken) {
...
 					}
-				};
-				cassandraOperation.updateRecord(Constants.SUNBIRD_KEY_SPACE_NAME, Constants.TABLE_ORG_HIERARCHY,
-						updateRequest, compositeKey);
-				response.getResult().put(Constants.ORGANIZATION_ID, orgId);
-				response.getResult().put(Constants.RESPONSE, Constants.SUCCESS);
+				} else {
+					updateRequest.put(Constants.PARENT_MAP_ID, Constants.SPV);
+				}
+				if (!StringUtils.isEmpty(mapId)){
```
karthik-tarento: This condition will be always true -- since mapId value assigned from createMapId method.... 
This suppose to check for null against requestData.get(Constants.MAP_ID))

src/main/java/org/sunbird/org/service/ExtendedOrgServiceImpl.java
```diff
@@ -208,16 +231,16 @@ private String validateOrgRequest(Map<String, Object> requestData) {
 
 		if (StringUtils.isEmpty((String) request.get(Constants.ORGANIZATION_TYPE))) {
```
karthik-tarento: use the orgType variable in this if condition too...

karthik-tarento (changes requested):

karthik-tarento:
Closing this PR since the version is increased to 4.8.0.
These code changes are committed from PR #169

PR #175, opened 2023-02-17
Offensive Data Report APIs

juhiagl8:


src/main/java/org/sunbird/assessment/service/OffensiveDataFlagServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+
+    @Override
+    public SBApiResponse updateFlag(String token, Map<String, Object> requestBody) {
+        String errMsg = "";
+        SBApiResponse response = ProjectUtil.createDefaultResponse(Constants.UPDATE_OFFENSIVE_DATA_FLAG);
+        String userId = RequestInterceptor.fetchUserIdFromAccessToken(token);
+        errMsg = validateUpdateRequest(requestBody, userId);
+        if (StringUtils.isEmpty(errMsg)) {
+            try {
+                if (StringUtils.isEmpty(errMsg)) {
```
karthik-tarento: why two if conditions

src/main/java/org/sunbird/assessment/service/OffensiveDataFlagServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+    @Autowired
+    CassandraOperation cassandraOperation;
+
+    @Override
+    public SBApiResponse createFlag(Map<String, Object> requestBody, String token) {
+        String errMsg = "";
+        String userId = RequestInterceptor.fetchUserIdFromAccessToken(token);
+        errMsg = validateAddRequest(requestBody, userId);
+        SBApiResponse response = ProjectUtil.createDefaultResponse(Constants.ADD_OFFENSIVE_DATA_FLAG);
+        try {
```
karthik-tarento: No need of try catch block for this method... insertRecord has it's own and it doesn't throw back.

src/main/java/org/sunbird/assessment/service/OffensiveDataFlagServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        return response;
+    }
+
+    @Override
+    public SBApiResponse getFlaggedData(String token) {
+        String userId = RequestInterceptor.fetchUserIdFromAccessToken(token);
+        String errMsg = "";
+        SBApiResponse response = ProjectUtil.createDefaultResponse(Constants.GET_OFFENSIVE_DATA_FLAG);
+        if (userId != null) {
+            try {
```
karthik-tarento: Do we need try catch for calling getRecordsByProperties method ?

src/main/java/org/sunbird/assessment/service/OffensiveDataFlagServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+                Map<String, Object> request = new HashMap<>();
+                request.put(Constants.USER_ID, userId);
+                List<Map<String, Object>> existingDataList = cassandraOperation.getRecordsByProperties(
+                        Constants.KEYSPACE_SUNBIRD, Constants.TABLE_OFFENSIVE_DATA_FLAGS, request, null);
+                if (existingDataList.isEmpty()) {
+                    errMsg = "Failed to update records into DB";
```
karthik-tarento: why are we treating it as error if there are no data ?

src/main/java/org/sunbird/assessment/service/OffensiveDataFlagServiceImpl.java
```diff
@@ -0,0 +1,214 @@
...
+        if (StringUtils.isNotBlank(errMsg)) {
+            response.getParams().setStatus(Constants.FAILED);
+            response.getParams().setErrmsg(errMsg);
+            response.setResponseCode(HttpStatus.BAD_REQUEST);
+            return response;
+        }
+        return response;
+    }
+
+    private String validateAddRequest(Map<String, Object> requestBody, String userId) {
```
karthik-tarento: can we merge the validateAddRequest and validateUpdateRequest methods ? just pass true / false flag to identify it is update or add method... based on that check contentStatus too...

src/main/java/org/sunbird/cassandra/utils/CassandraOperationImpl.java
```diff
@@ -225,8 +227,10 @@ public Map<String, Object> updateRecord(String keyspaceName, String tableName, M
 			session.execute(updateQuery);
 			response.put(Constants.RESPONSE, Constants.SUCCESS);
 		} catch (Exception e) {
-			logger.error(e.getMessage());
-			throw e;
```
karthik-tarento: Adding more details in response object is fine... but do not change the behaviour without any analysis... you have removed "throw e" -- that is going affect existing implementation

src/main/resources/cassandratablecolumn.properties
```diff
@@ -43,4 +43,12 @@ filename=fileName
...
\ No newline at end of file
+sborgid=sbOrgId
+contexttype=contextType
+contexttypeid=contextTypeId
+acquiredchannel=acquiredChannel
+additionalparams=additionalParams
+requestedby=requestedBy
+contextstatus=contextStatus
+createdat=createdAt
+updatedat=updatedAt
```
karthik-tarento: can we use createdDate and updatedDate - existing values instead of new ones?
since this is used only in backend - no need to change the API details ??

karthik-tarento (changes requested):

karthik-tarento (approved):

PR #182, opened 2023-03-27
new api's for getting master data's

Manas-tarento:


src/main/java/org/sunbird/common/util/Constants.java
```diff
@@ -35,6 +35,21 @@ public class Constants {
 	public static final String DEPARTMENT_NAME = "department_name";
 	public static final String UNIT_NAME = "unit_name";
 	public static final String LANGUAGES_KNOWN = "languages_known";
+	public static final String LANGUAGE = "language";
+	public static final String LANGUAGES = "languages";
+	public static final String GRADUATION = "graduation";
+	public static final String GRADUATIONS = "graduations";
+	public static final String POST_GRADUATION = "postGraduation";
+	public static final String POST_GRADUATIONS = "postGraduationS";
```
karthik-tarento: typo in last char

src/main/java/org/sunbird/searchby/model/MasterData.java
```diff
@@ -0,0 +1,54 @@
+package org.sunbird.searchby.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class MasterData {
+    private String Id;
```
karthik-tarento: variable name shouldn't start in capital

src/main/java/org/sunbird/searchby/model/MasterData.java
```diff
@@ -0,0 +1,54 @@
...
+    private String contextType;
+    @JsonProperty("contextname")
```
karthik-tarento: No need to add jsonProperty -- you need to add contextname to contextName conversion in cassandratablecolumn.properties file...

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
```diff
@@ -0,0 +1,366 @@
...
+import java.util.*;
+
+@Service
+public class MasterDataServiceImpl implements MasterDataService {
+
+    private CbExtLogger logger = new CbExtLogger(getClass().getName());
+    @Autowired
+    CbExtServerProperties cbExtServerProperties;
+    @Autowired
+    RedisCacheMgr redisCacheMgr;
```
karthik-tarento: No need to use redisCacheMgr - we can add Jenkins cache for these if required.

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
```diff
@@ -0,0 +1,366 @@
...
+                response.setResponseData(positionMap.get(Constants.POSITIONS_CACHE_NAME));
+            }
+        } catch (Exception e) {
+            response.getStatusInfo().setStatusCode(HttpStatus.INTERNAL_SERVER_ERROR.value());
+        }
+
+        return response;
+    }
+
+    private Map<String, List<MasterData>> updateDesignationDetails(String authUserToken) throws Exception {
```
karthik-tarento: why this method is called as updateDesignationDetails ? this method is not required and these implementation can be moved into getListPositions

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
```diff
@@ -0,0 +1,366 @@
...
+            logger.info("Exception occurred while fetching details from the database");
+            errMsg = "Exception occurred while fetching details from the database";
+        }
+        if (StringUtils.isNotBlank(errMsg)) {
+            response.put(Constants.ERROR_MESSAGE, errMsg);
+            response.put(Constants.RESPONSE_CODE, HttpStatus.BAD_REQUEST);
+            return response;
+        }
+        Map<String,Object> result = constructResponse(listOfMasterData, type);
+        response.put(Constants.RESPONSE_CODE, HttpStatus.OK);
```
karthik-tarento: Are we using the extra information added in the response object ??
Controller is reading only - result information added in response ?? simply send only map (i.e. return response from constructResponse) object back to controller instead of response object.

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
```diff
@@ -0,0 +1,366 @@
...
+        response.put(Constants.RESULT, result);
+        return response;
+    }
+
+    private Map<String, Object> constructResponse(List<Map<String, Object>> listOfMasterData, String contextType) {
```
karthik-tarento: If we have contextType as same value which match against return response -- then no need of this method ?
if so -- go ahead and use the contexttype as same as how you are going to add send response object.

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
```diff
@@ -0,0 +1,366 @@
...
+                        && !Constants.SUCCESS.equalsIgnoreCase((String) updateResponse.get(Constants.RESPONSE))) {
+                    errMsg = String.format("Failed to update details");
+                    response.getParams().setErrmsg(errMsg);
+                    return response;
+                }
+            } else {
+                Map<String, Object> propertyMap = new HashMap<>();
+                propertyMap.put(Constants.CONTEXT_TYPE.toLowerCase(), masterData.get(Constants.CONTEXT_TYPE));
+                List<Map<String, Object>> totalMasterData = cassandraOperation.getRecordsByProperties(Constants.KEYSPACE_SUNBIRD, Constants.TABLE_MASTER_DATA, propertyMap, new ArrayList<>());
+                long currentMaxId = totalMasterData.size();
```
karthik-tarento: Unnecessary call again to DB... you already retrieved the listOfMasterData for specific contextType -- get the size from that and add +1

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
```diff
@@ -0,0 +1,366 @@
...
+        }
+        if (StringUtils.isEmpty((String) masterData.get(Constants.CONTEXT_TYPE))) {
+            params.add(Constants.CONTEXT_TYPE);
+        }
+        if (ObjectUtils.isEmpty(masterData.get(Constants.CONTEXT_NAME))) {
+            params.add(Constants.CONTEXT_NAME);
+        }
+        if (ObjectUtils.isEmpty(masterData.get(Constants.CONTEXT_DATA))) {
+            params.add(Constants.CONTEXT_DATA);
+        }
```
karthik-tarento: contextData is not mandatory for all

karthik-tarento (changes requested):

src/main/java/org/sunbird/common/util/Constants.java
Manas-tarento: Done

src/main/java/org/sunbird/searchby/model/MasterData.java
Manas-tarento: Done

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: Ok Removed redis

src/main/java/org/sunbird/searchby/model/MasterData.java
Manas-tarento: change the name

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: Moved

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: the names means the type was different in db singular plural that's why

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: All will be plural is it ok ?

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: ok will removed

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: db name changed

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: db col name changed

src/main/java/org/sunbird/searchby/service/MasterDataServiceImpl.java
Manas-tarento: as we discuss here no change required

PR #200, opened 2023-04-12
changes for positions

juhiagl8:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -209,6 +211,9 @@ private List<String> validateEmailContactAndDomain(UserRegistration userRegistra
         if (!ProjectUtil.validateContactPattern(userRegistration.getPhone())) {
             errList.add("Invalid Contact Number");
         }
+        if (!userUtilityService.validatePosition(userRegistration.getPosition())) {
+            errList.add("Invalid Position");
+        }
```
karthik-tarento: You need to check the position is null or not first...
if the position is null - the current code will still try to call DB -- which will throw error.. 
as mentioned previous - if position is null - then error should be - Positions is missing.

karthik-tarento (changes requested):

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -200,15 +200,18 @@ private List<String> validateEmailContactAndDomain(UserRegistration userRegistra
...
         }
         if (!ProjectUtil.validateContactPattern(userRegistration.getPhone())) {
             errList.add("Invalid Contact Number");
         }
+        if (StringUtils.isBlank(userRegistration.getPosition())) {
+            errList.add("Position is missing");
+        }
+        if (!userUtilityService.validatePosition(userRegistration.getPosition())) {
+            errList.add("Invalid Position");
+        }
```
karthik-tarento: it should else block --> if position is not null then only you have to validate, right ?

karthik-tarento (changes requested):

PR #206, opened 2023-04-25
added changes for position

juhiagl8:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -209,6 +211,19 @@ private List<String> validateEmailContactAndDomain(UserRegistration userRegistra
...
+        if (StringUtils.isBlank(userRegistration.getPosition())) {
+            errList.add("Position is missing");
+        } else {
+            if (!userUtilityService.validatePosition(userRegistration.getPosition())) {
+                errList.add("Invalid Position");
+            }
+        }
+        if (!StringUtils.isBlank(userRegistration.getTag())) {
+            errList.add("Tag is missing");
+        }
```
karthik-tarento: Just be careful on using negation.

src/main/java/org/sunbird/common/util/Constants.java
```diff
@@ -666,7 +666,9 @@ public class Constants {
 	public static final String SUCCESSFUL_RECORDS_COUNT = "successfulRecordsCount";
 	public static final String FAILED_RECORDS_COUNT = "failedRecordsCount";
 	public static final String API_FILE_DOWNLOAD = "api.file.download";
-	private Constants() {
+    public static final String TAG = "tag";
+
+    private Constants() {
```
karthik-tarento: why the alignments are too bad?

karthik-tarento (changes requested):

PR #218, opened 2023-05-11
changes for empty fields

juhiagl8:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -118,20 +118,37 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
                     rowIterator.next();
                 }
                 while (rowIterator.hasNext()) {
+                    StringBuffer str = new StringBuffer();
+                    List<String> errList = new ArrayList<>();
                     Row nextRow = rowIterator.next();
-                    if (StringUtils.isBlank(nextRow.getCell(0).getStringCellValue())) {
+                    if (totalRecordsCount > 0 && nextRow.getCell(0) == null) {
```
karthik-tarento: What if cell is not null but it has empty value ??

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -141,21 +158,30 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
                     }
                     totalRecordsCount++;
-                    if (errList.isEmpty()) {
-                        boolean isUserCreated = userUtilityService.createUser(userRegistration);
-                        if (isUserCreated) {
-                            noOfSuccessfulRecords++;
-                            statusCell.setCellValue(Constants.SUCCESS.toUpperCase());
-                            errorDetails.setCellValue("");
+                    if (!errList.isEmpty()) {
+                        str.append("Failed to Process the Uploaded File. Error Details - [").append(errList).append("]");
```
karthik-tarento: This is not the code I've expected. errList still contains "is missing" for each field. 
Just to make sure the error message is as follows:
Failed to process user record. Missing Parameters - [First Name, Last Name, Email, Phone]

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -141,21 +158,30 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
+                        failedRecordsCount++;
+                        statusCell.setCellValue(Constants.FAILED.toUpperCase());
```
karthik-tarento: Can we use different constant which has all upper case instead of calling toUpperCase

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -141,21 +158,30 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
+                        errorDetails.setCellValue(errList.toString());
```
karthik-tarento: use the str variable instead of errList

karthik-tarento (changes requested):

PR #222, opened 2023-05-16
changes for verified karmayogi badge

juhiagl8:


src/main/java/org/sunbird/common/util/CbExtServerProperties.java
```diff
@@ -1617,4 +1620,12 @@ public Integer getOrgSearchResponseDefaultLimit() {
...
 		this.orgSearchResponseDefaultLimit = orgSearchResponseDefaultLimit;
 	}
+
+	public String getLmsSystemSettingsVerifiedProfileFieldsPath() {
+		return verifiedProfileFieldsPath;
+	}
+
+	public void setLmsSystemSettingsVerifiedProfileFieldsPath(String verifiedProfileFieldsPath) {
+		this.verifiedProfileFieldsPath = verifiedProfileFieldsPath;
+	}
```
karthik-tarento: the getter setter method name must be driven from variable names

karthik-tarento (changes requested):

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -177,11 +173,17 @@ public SBApiResponse profileUpdate(Map<String, Object> request, String userToken
 						getModifiedPersonalDetails(profileDetailsMap.get(changedObj), requestData);
 					}
 				}
+				log.info("Existing Profile Details : " + new Gson().toJson(existingProfileDetails));
+				String schema = getVerifiedProfileSchema();
+				if (validateJsonAgainstSchema(schema, new Gson().toJson(existingProfileDetails))) {
```
karthik-tarento: toJson -- is this throws error ?? if so, just pass the existingProfileDetails to the method and convert it to json inside a try catch in that method...

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -305,6 +307,31 @@ public SBApiResponse profileUpdate(Map<String, Object> request, String userToken
 		return response;
 	}
 
+	public boolean validateJsonAgainstSchema(String jsonSchema, String jsonData) {
```
karthik-tarento: this method should have one try catch not three....

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -136,6 +131,7 @@ public SBApiResponse profileUpdate(Map<String, Object> request, String userToken
 			}
 			Map<String, Object> responseMap = userUtilityService.getUsersReadData(userId, StringUtils.EMPTY,
 					StringUtils.EMPTY);
+			log.info("Reading Profile Details : " + responseMap.toString());
```
karthik-tarento: Remove this info log

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -177,11 +173,17 @@ public SBApiResponse profileUpdate(Map<String, Object> request, String userToken
 						getModifiedPersonalDetails(profileDetailsMap.get(changedObj), requestData);
 					}
 				}
+				log.info("Existing Profile Details : " + new Gson().toJson(existingProfileDetails));
```
karthik-tarento: Remove this info logn

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -177,11 +173,17 @@ public SBApiResponse profileUpdate(Map<String, Object> request, String userToken
...
+				String schema = getVerifiedProfileSchema();
```
karthik-tarento: schema is not useful in this method... it is required only inside valideJsonAgainstSchema... 
just call the getVerifiedProfileSchema inside that method and avoid passing the value...

karthik-tarento (changes requested):

karthik-tarento (approved):

PR #236, opened 2023-05-31
changes for invalid tags

juhiagl8:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -212,7 +212,10 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
                         setErrorDetails(str, errList, statusCell, errorDetails);
                         failedRecordsCount++;
                     } else {
-                        invalidErrList = validateEmailContactAndDomain(userRegistration);
+                        List<String> err = validateEmailContactAndDomain(userRegistration);
```
karthik-tarento: No need to create err list object... 
just call --> invalidErrList.addAll(validateEmailContactAndDomain(userRegistration));

karthik-tarento (changes requested):

PR #257, opened 2023-06-27
Group changes 4.8.3

juhiagl8:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -141,57 +141,61 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
                     }
                     if (nextRow.getCell(2) == null || nextRow.getCell(2).getCellType() == CellType.BLANK) {
                         errList.add("Phone");
                     } else {
                         if (nextRow.getCell(2).getCellType() == CellType.NUMERIC) {
                             phone = NumberToTextConverter.toText(nextRow.getCell(2).getNumericCellValue());
-                            userRegistration.setPhone(phone);
+                            userRegistration.setPhone(phone.trim());
+                        }
+                        else if (nextRow.getCell(2).getCellType() == CellType.STRING)
```
karthik-tarento: what if cellType is not numeric / String ??
need to throw error in that case - so that we know why it is failed...

src/main/java/org/sunbird/user/service/UserUtilityServiceImpl.java
```diff
@@ -636,6 +633,7 @@ public String updateBulkUploadUser(UserRegistration userRegistration) {
 			additionalProperties.put(Constants.EXTERNAL_SYSTEM, userRegistration.getExternalSystem());
 		}
 		profileDetails.put(Constants.ADDITIONAL_PROPERTIES, additionalProperties);
+		profileDetails.put(Constants.VERIFIED_KARMAYOGI, false);
```
karthik-tarento, line 636: is the mandatoryFieldsExists field added with false ??

karthik-tarento (changes requested):

karthik-tarento (approved):

PR #276, opened 2023-08-22
Dev #96252 Provide API to update attendance manually for offline session

Sahil-tarento:
Dev #96252 Provide API to update attendance manually for offline session


SaipradeepR (changes requested):
Cancel this PR & create new PR with single commit & appropriate comments.

PR #375, opened 2023-12-22
Cb Plan API changes for populating user Data Info

Sahil-tarento:


src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
```diff
@@ -922,4 +936,31 @@ private String validateListAPI(String orgId, SunbirdApiRequest request, CbPlanSe
 
         return errMsg;
     }
+
+    private String getDesignationForUser(String profileDetails) throws IOException {
```
karthik-tarento: don't throw the exception... just log and return empty string...

src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
```diff
@@ -89,9 +89,17 @@ public SBApiResponse createCbPlan(SunbirdApiRequest request, String userOrgId, S
             try {
                 requestMap.put(Constants.DRAFT_DATA, mapper.writeValueAsString(cbPlanDto));
                 requestMap.put(Constants.STATUS, Constants.DRAFT);
-                cassandraOperation.insertRecord(Constants.KEYSPACE_SUNBIRD, Constants.TABLE_CB_PLAN, requestMap);
-                response.getResult().put(Constants.ID, cbPlanId);
+                SBApiResponse resp = cassandraOperation.insertRecord(Constants.KEYSPACE_SUNBIRD, Constants.TABLE_CB_PLAN, requestMap);
+                if (!resp.get(Constants.RESPONSE).equals(Constants.SUCCESS)) {
```
karthik-tarento: always check against constant instead of object or variable... avoid negation as much as possible.. 
if (Constants.SUCCESS.equals(resp.get(Constants.RESPONSE))) {
	response.getResult().put(Constants.STATUS, Constants.CREATED);
        response.getResult().put(Constants.ID, cbPlanId);
} else {
    response.getParams().setStatus(Constants.FAILED);
    response.getParams().setErrmsg("Failed to Create CB Plan for OrgId: " + userOrgId + " message: " + resp.getParams().getErrmsg());
    response.setResponseCode(HttpStatus.INTERNAL_SERVER_ERROR);
}

karthik-tarento (changes requested):

src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
Sahil-tarento: Made the implementation

src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
Sahil-tarento: Made the implementation

karthik-tarento (approved):

PR #391, opened 2024-01-02
added API for cbplan request content

ravisaurav-tarento:
HI @karthik-tarento kindly review this PR

### DIFF
--- a/src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
+++ b/src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
@@ -515,20 +515,16 @@ public class CbPlanServiceImpl implements CbPlanService {
         Map<String, Object> competency = (Map<String, Object>) contentRequest.get(Constants.COMPETENCY);
         List<String> providersOrgId = (List<String>) contentRequest.get("providerList");
         String description = (String) contentRequest.get(Constants.DESCRIPTION);
-        boolean hasError = false;
         try {
             String userId = validateAuthTokenAndFetchUserId(token);
-            hasError = validateRequestCbplanPayload(userId, competency, providersOrgId, response);
-            if(hasError)
+            if (validateRequestCbplanPayload(userId, competency, providersOrgId, response))
                 return response;
             Map<String, String> mdoInfo = userUtilityService.getUserDetails(Collections.singletonList(userId), new ArrayList<>()).get(userId);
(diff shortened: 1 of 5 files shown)


src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
```diff
@@ -509,6 +509,37 @@ public SBApiResponse getCBPlanListForUser(String userOrgId, String authUserToken
         return response;
     }
 
+    public SBApiResponse requestCbplanConten(SunbirdApiRequest request, String userOrgId, String token){
```
karthik-tarento: Follow the existing implementation style.
check userId is not empty - if empty return 400 error
have a validation method to check competency and providerList is not null from the request - if either is null return 400 error

src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
```diff
@@ -509,6 +509,37 @@ public SBApiResponse getCBPlanListForUser(String userOrgId, String authUserToken
...
+        try{
+            String userId = validateAuthTokenAndFetchUserId(token);
+            Map<String, String> mdoInfo = userUtilityService.getUserDetails(Collections.singletonList(userId), new ArrayList<>()).get(userId);
+            String mdoOrgId = mdoInfo.get(Constants.ROOT_ORG_ID);
+            String competencyJsonString = mapper.writeValueAsString(competency);
+
+            Map<String, Object> propertiesMap = new HashMap<>();
+            propertiesMap.put(Constants.ORG_ID, mdoOrgId);
+            propertiesMap.put(Constants.ID, UUID.randomUUID().toString());
+            propertiesMap.put(Constants.COMPETENCY_INFO , Collections.singletonList(competencyJsonString));
```
karthik-tarento: can't we simply use competency list object ?

karthik-tarento (changes requested):

src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
ravisaurav-tarento: This is implemented please verify.

src/main/java/org/sunbird/cbp/service/CbPlanServiceImpl.java
ravisaurav-tarento: This is updated, we are storing value as an object.

PR #458, opened 2024-02-21
 KB-2485 - BE | Dev | Create a new API for downloading a password-protected zip report

tarentomaheshvakkund:


src/main/java/org/sunbird/operationalreports/service/OprationalReportServiceImpl.java
SaipradeepR, line 1: 1.Transfer the string literals "/unzippath" , "%s/%s/%s/%s",  "/" to a constants file.
2.While generating a file at the temporary location, the lack of a unique identification posed a potential problem when multiple download requests occurred simultaneously from different MDOs. To address this issue, consider appending a unique ID each time a download is made to the local temporary folder.

SaipradeepR: 3. remove unnecessary comments like before log statements 
eg.     // Log success message
            logger.info("Zip folder unlocked successfully.");

SaipradeepR (changes requested):

src/main/java/org/sunbird/operationalreports/service/OprationalReportServiceImpl.java
tarentomaheshvakkund: 1. Added the constant in the Constants.java file in the cb-ext service.
2. Added a uuid as paramter of the output file creation to maintain uniqueness while processing the request.
3. Removed the unnecessary comments before the logs.

PR #480, opened 2024-03-05
KB-2925 | New API for the OTP generation for email/phone verification

tarentomaheshvakkund:
1. Added a new version of extPatch API.


src/main/resources/application.properties
```diff
@@ -390,3 +390,4 @@ report.property.file.allMdo=ContentCompetencyMapping.csv
 operational.reports.passwordlength=15
 operational.reports.unzip.password=123456
 es.default.result.limit=250
+secret.key.primaryEmail.validation=secretKey
```
karthik-tarento: This is not primaryEmail validation... it's extPatch operation validation.
rename the variable to -> secretKeyTokenValidation

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -1931,6 +1939,74 @@ public List<String> adminApprovalFields() {
...
+	 * @param rootOrgId The root organization ID.
+	 * @return SBApiResponse object containing the response of the profile update.
+	 * @throws Exception if any error occurs during the profile update process.
+	 */
+	@Override
+	public SBApiResponse profileUpdateV2(Map<String, Object> request, String userToken, String authToken, String rootOrgId) throws Exception {
+		// Extracting the request map from the input request
+		Map<String, Object> requestMap = (Map<String, Object>) request.get(Constants.REQUEST);
+		// Validating the token from the request map
+		if (validateToken((String) requestMap.get("token"))) {
```
karthik-tarento: This method throws exception back to controller - which is wrong
Also it should not return null.

Refer the other method implementations - service methods should return only SBApiResponse -- with OK or with error messages.

src/main/java/org/sunbird/profile/exception/ProfileRequestException.java
```diff
@@ -0,0 +1,22 @@
+package org.sunbird.profile.exception;
+
+/**
+ * @author mahesh.vakkund
+ */
+public class ProfileRequestException extends Exception {
```
karthik-tarento: No need of this new exception class

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -1931,6 +1939,74 @@ public List<String> adminApprovalFields() {
...
+			// Parsing the token and extracting its subject (claims)
+			String tokens = Jwts.parser().setSigningKey(serverConfig.getSecretKeyPrimaryEmailValidation()).parseClaimsJws(token).getBody().getSubject();
+			// Parsing the subject (claims) into a map using Jackson ObjectMapper
+			Map<String, String> parsedMap = mapper.readValue(tokens, new TypeReference<Map<String, String>>() {
+			});
+			// Checking if the 'contextType' key exists and is not empty
+			return parsedMap.get(Constants.CONTEXT_TYPE) != null && !parsedMap.get(Constants.CONTEXT_TYPE).isEmpty();
+		} catch (ExpiredJwtException e) {
+			// Handling ExpiredJwtException
+			throw new ProfileRequestException("ExpiredJwt exception has occurred: " + e.getMessage());
```
karthik-tarento: throw new Exception(errMsg, e) -- should be fine for all of this exceptions

karthik-tarento (changes requested):

src/main/resources/application.properties
tarentomaheshvakkund: 1. changed to secret.key.secretKeyTokenValidation

src/main/java/org/sunbird/profile/exception/ProfileRequestException.java
tarentomaheshvakkund: 1.Deleted the class for custom exception

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
tarentomaheshvakkund: 1.Handled with try and catch and return the SBIResponse with error message.
2. Removed the exception sent to the controller class.

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
tarentomaheshvakkund: 1. It would throw Exception with the error message.

karthik-tarento (approved):

PR #486, opened 2024-03-07
KB-3119 | Enhancement of the  API for the OTP generation for email/phone verification

tarentomaheshvakkund:
1. Enhancement of the API to check if the email/phone number to be updated already exists.

### DIFF
--- a/src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
+++ b/src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
@@ -134,17 +134,13 @@ public class ProfileServiceImpl implements ProfileService {
 				return response;
 			}
 			Map<String, Object> profileDetailsMap = (Map<String, Object>) requestData.get(Constants.PROFILE_DETAILS);
-			String validatePhoneEmail = validateExistingPhoneEmail(profileDetailsMap);
-            switch (validatePhoneEmail) {
-                case Constants.EMAIL_EXIST_ERROR:
-                case Constants.PHONE_NUMBER_EXIST_ERROR:
-                    response.setResponseCode(HttpStatus.BAD_REQUEST);
-                    response.getParams().setStatus(Constants.FAILED);
-                    response.getParams().setErrmsg("Email Id or Phone Number Already Exists");
-                    return response;
-                default:
-                    break;
-            }
+			String validatePhoneEmailErrMsg = validateExistingPhoneEmail(profileDetailsMap);
+			if (!validatePhoneEmailErrMsg.isEmpty()) {
+				response.setResponseCode(HttpStatus.BAD_REQUEST);
+				response.getParams().setStatus(Constants.FAILED);
+				response.getParams().setErrmsg(validatePhoneEmailErrMsg);
+				return response;
+			}
 			List<String> approvalFieldList = approvalFields();
 			String newDeptName = checkDepartment(profileDetailsMap);
 			Map<String, Object> transitionData = new HashMap<>();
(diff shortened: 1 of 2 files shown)


src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -2019,6 +2026,27 @@ public String validateToken(String token) throws Exception {
...
+	 *
+	 * @param profileDetailsMap A map containing the profile details.
+	 * @return Constants.OK if neither email nor phone number exists, Constants.EMAIL_EXIST_ERROR if email already exists,
+	 *         Constants.PHONE_NUMBER_EXIST_ERROR if phone number already exists.
+	 */
+	private String validateExistingPhoneEmail(Map<String, Object> profileDetailsMap) {
+		// Extract personal details from the profileDetailsMap
+		Map<String, Object> personalDetails = (Map<String, Object>) profileDetailsMap.get(Constants.PERSONAL_DETAILS);
+		// Check if the provided email already exists in the system
+		if (userUtilityService.isUserExist(Constants.EMAIL, (String) personalDetails.get(Constants.EMAIL))) {
```
karthik-tarento: There is a possibility that profileDetails object doesn't have personalDetails... 
Do check personalDetails exist -- also check personalDetails contains key -> Email -- if so then call isUserExist method

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -2019,6 +2026,27 @@ public String validateToken(String token) throws Exception {
...
+			return Constants.EMAIL_EXIST_ERROR; // Return error message if email already exists
+		}
+		// Check if the provided phone number already exists in the system
+		if (userUtilityService.isUserExist(Constants.PHONE, (String) personalDetails.get(Constants.PHONE))) {
```
karthik-tarento: There is a possibility that profileDetails object doesn't have personalDetails... 
Do check personalDetails exist -- also check personalDetails contains key -> Phone -- if so then call isUserExist method

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -134,6 +134,13 @@ public SBApiResponse profileUpdate(Map<String, Object> request, String userToken
 				return response;
 			}
 			Map<String, Object> profileDetailsMap = (Map<String, Object>) requestData.get(Constants.PROFILE_DETAILS);
+			String validatePhoneEmail = validateExistingPhoneEmail(profileDetailsMap);
+			if (validatePhoneEmail.equals(Constants.EMAIL_EXIST_ERROR) || validatePhoneEmail.equals(Constants.PHONE_NUMBER_EXIST_ERROR)) {
```
karthik-tarento: validatePhoneEmail -- rename to errMsg 
Just check string is not empty... if not empty return bad_request...

karthik-tarento (changes requested):

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
tarentomaheshvakkund: have added check to validate the key should be present else it will return appropriate message

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
tarentomaheshvakkund: have added check to validate the key should be present else it will return appropriate message

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
tarentomaheshvakkund: handled with validation.

PR #487, opened 2024-03-07
API for Read access & changes in Grant Access API

dkttarento:
API for Read access & changes in Grant Access API


src/main/java/org/sunbird/operationalreports/controller/OperationalReportController.java
```diff
@@ -39,4 +39,13 @@ public ResponseEntity<InputStreamResource> downloadFile(
             @PathVariable String fileName) throws IOException {
         return operationalReport.downloadFile(reportType, date,orgId,fileName);
     }
+
+    @GetMapping("admin/readaccess")
+        public ResponseEntity<SBApiResponse> readAccess(
+                @RequestHeader(Constants.X_AUTH_TOKEN) String authToken,
+                @RequestHeader(Constants.X_AUTH_USER_ORG_ID) String userOrgId) throws Exception {
```
karthik-tarento, line 46: get the orgId from user access token... 
from user access token - get the userId
from userId -- read the cassandra table sunbird.user -- for given user id... 
from user record - get the rootOrgId

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
-                Map<String, Object> userSearchResponse = (Map<String, Object>) map.get(Constants.RESPONSE);
-                List<Map<String, Object>> contents = (List<Map<String, Object>>) userSearchResponse.get(Constants.CONTENT);
-                Set<String> rootOrgIds = new HashSet<>();
-                String rootOrgId = null;
-                if (!CollectionUtils.isEmpty(contents)) {
-                    for (Map<String, Object> content : contents) {
-                        rootOrgId = (String) content.get(Constants.ROOT_ORG_ID);
-                        rootOrgIds.add(rootOrgId);
-                    }
+        while (true) {
```
karthik-tarento: while loop is unnecessary for this use case. Do not use while loop

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
+            try {
+                Map<String, Object> mdoAdminDetails = (Map<String, Object>) request.getRequest();
+                String mdoAdminUserId = (String) mdoAdminDetails.get(Constants.USER_ID);
+                String reportExpiryDateRequest = (String) mdoAdminDetails.get(Constants.REPORT_EXPIRY_DATE);
+                Map<String, String> headersValue = new HashMap<>();
+                headersValue.put(Constants.CONTENT_TYPE, Constants.APPLICATION_JSON);
+                String mdoLeaderUserId = accessTokenValidator.fetchUserIdFromAccessToken(authToken);
+                if (null == mdoLeaderUserId) {
+                    throw new HttpServerErrorException(HttpStatus.INTERNAL_SERVER_ERROR, "UserId Couldn't fetch from auth token");
```
karthik-tarento: this should be 1st statement... the above statements are useless if the userId is invalid... so do bring those statements down... 

String userIdFromToken = accessTokenValidator.fetchUserIdFromAccessToken(userToken);
			if (StringUtils.isBlank(userIdFromToken)) {
				response.setResponseCode(HttpStatus.BAD_REQUEST);
				response.getParams().setStatus(Constants.FAILED);
				response.getParams().setErrmsg("Invalid UserId in the request");
				return response;
			}

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
+                    String rootOrgId = null;
+                    if (!CollectionUtils.isEmpty(contents)) {
+                        for (Map<String, Object> content : contents) {
+                            rootOrgId = (String) content.get(Constants.ROOT_ORG_ID);
+                            rootOrgIds.add(rootOrgId);
+                        }
+                    }
+                    if (rootOrgIds.size() != 1) {
+                        logger.error("Failed to grant access due to different org.");
+                        throw new HttpClientErrorException(HttpStatus.BAD_REQUEST, "Failed to grant access due to different org.");
```
karthik-tarento, line 126: Do not throw exception - construct response object with error code and return the response object

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
+                    List<Map<String, Object>> contents = (List<Map<String, Object>>) userSearchResponse.get(Constants.CONTENT);
+                    Set<String> rootOrgIds = new HashSet<>();
+                    String rootOrgId = null;
+                    if (!CollectionUtils.isEmpty(contents)) {
+                        for (Map<String, Object> content : contents) {
+                            rootOrgId = (String) content.get(Constants.ROOT_ORG_ID);
+                            rootOrgIds.add(rootOrgId);
+                        }
+                    }
+                    if (rootOrgIds.size() != 1) {
```
karthik-tarento, line 124: Need to check that mdoUserOrgId -- matches against the mdoAdminOrgId value -- if not return error response.

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
-                List<String> assignedRoles = new ArrayList<>();
-                for (Map<String, Object> roleMap : roles) {
-                    assignedRoles.add((String) roleMap.get(Constants.ROLE));
-                }
-                if (!assignedRoles.contains(Constants.MDO_ADMIN)) {
-                    throw new HttpClientErrorException(HttpStatus.BAD_REQUEST, "The Grantee doesn't have MDO_ADMIN role");
+                logger.info("MDO Leader User ID : {}", mdoLeaderUserId);
+                StringBuilder url = new StringBuilder(configuration.getLmsServiceHost()).append(configuration.getLmsUserSearchEndPoint());//call sunbird user table =>rootorg set against the incoming rqst body.
+                Map<String, Object> userSearchRequest = getSearchObject(Arrays.asList(mdoLeaderUserId, mdoAdminUserId));
+                Map<String, Object> searchProfileApiResp = outboundReqService.fetchResultUsingPost(url.toString(), userSearchRequest, headersValue);
```
karthik-tarento: use the existing methods in userutilityserviceimpl -- getUserDetailsFromDB method can be used here... 
do not construct request and process the response again -- which is already done inside userutilityserviceimpl

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
+                    }
+                    if (rootOrgIds.size() != 1) {
+                        logger.error("Failed to grant access due to different org.");
+                        throw new HttpClientErrorException(HttpStatus.BAD_REQUEST, "Failed to grant access due to different org.");
+                    }
+                    Map<String, Object> mdoAdminData = userUtilityService.getUsersReadData(mdoLeaderUserId, null, null);
+                    List<String> mdoAdminRoles = (List<String>) mdoAdminData.get(Constants.ROLES);
+                    if (!mdoAdminRoles.contains(Constants.MDO_ADMIN)) {
+                        throw new HttpClientErrorException(HttpStatus.BAD_REQUEST, "The Grantee doesn't have MDO_ADMIN role");
+                    }
```
karthik-tarento: is this suppose to be mdoAdminUserId -- which is read from request body... ?
why this is mdoLeaderUserId -- which is read from access token?

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -73,82 +76,96 @@ public class OperationalReportServiceImpl implements OperationalReportService {
...
+                    Map<String, Object> assignRoleReq = new HashMap<>();
+                    Map<String, Object> roleRequestBody = new HashMap<>();
+                    roleRequestBody.put(Constants.ORGANIZATION_ID, rootOrgId);
+                    roleRequestBody.put(Constants.USER_ID, mdoAdminUserId);
+                    roleRequestBody.put(Constants.ROLES, mdoAdminRoles);
+                    assignRoleReq.put(Constants.REQUEST, roleRequestBody);
+                    outboundRequestHandlerService.fetchResultUsingPost(serverConfig.getSbUrl() + serverConfig.getSbAssignRolePath(), assignRoleReq,
+                            headersValue);
+                    Map<String, Object> primaryKeyMap = new HashMap<>();
+                    primaryKeyMap.put(Constants.USER_ID_LOWER, mdoAdminUserId);
```
karthik-tarento, line 154: No need of these lines -- just call the private method which is used to update the record.

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -359,14 +376,45 @@ private Map<String, Object> getSearchObject(List<String> userIds) {
         return requestObject;
     }
 
-    private long getParsedDate(String reportExpiryDate) throws ParseException {
-        long reportExpiryDateMillis;
-        SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd");
-        Date parsedDate = dateFormat.parse(reportExpiryDate);
-        reportExpiryDateMillis = parsedDate.getTime();
-        return reportExpiryDateMillis;
+    public SBApiResponse readGrantAccess(String authToken,String userOrgId) {
```
karthik-tarento, line 387: userOrgId -- should not be coming as input from controller... 

read userId from access token, call userutilityserviceImpl.getUserDetailsFromDB -- from the user record get the rootOrgId value...

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -359,14 +376,45 @@ private Map<String, Object> getSearchObject(List<String> userIds) {
...
+        }
+        response.getResult().put(Constants.STATUS, Constants.SUCCESS);
+        response.getResult().put(Constants.RESPONSE, result);
+        return response;
+    }
+
+    private Map<String, Object> getReportAccessDetails(String userOrgId) {
+        Map<String, Object> primaryKeyMap = new HashMap<>();
+        primaryKeyMap.put(Constants.USER_ORG_ID, userOrgId);
+        return cassandraOperation.getRecordsByPropertiesByKey(Constants.KEYSPACE_SUNBIRD,
```
karthik-tarento, line 407: use the cassandraOperation.***WithoutAllowFiltering -- methods should be used...

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -359,14 +376,45 @@ private Map<String, Object> getSearchObject(List<String> userIds) {
...
+        Map<String, Object> primaryKeyMap = new HashMap<>();
+        primaryKeyMap.put(Constants.USER_ORG_ID,userOrgId);
+        Map<String, Object> userAccessReportExpiryDetails = getReportAccessDetails(userOrgId);
+        List<Map<String, Object>> result = new ArrayList<>();
+        for(Map.Entry<String, Object> entry:userAccessReportExpiryDetails.entrySet()){
+            Map<String, Object> mdoIdAccessDetails = new HashMap<>();
+            mdoIdAccessDetails.put(Constants.USER_ID,entry.getKey());
+            mdoIdAccessDetails.put(Constants.REPORT_ACCESS_EXPIRY, ((Map<String, Object>) entry.getValue()).get(Constants.USER_REPORT_EXPIRY_DATE));
+            result.add(mdoIdAccessDetails);
+        }
```
karthik-tarento, line 398: do we need to convert the object again ?
why can't we simply add userAccessReportExpireDetails map in the result ??
make sure cassandratablecolumnmapping.properties is updated with necessary conversion to do this... 

what will happen if the value is empty ?

karthik-tarento (changes requested):

PR #492, opened 2024-03-08
KB-3163 | BE | Dev | Create a new API to get the File Information of the file stored in the s3 bucket

tarentomaheshvakkund:
1. Created a new API to get the file information of the file stored in the s3 bucket.

### DIFF
--- a/src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
+++ b/src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
@@ -382,20 +382,21 @@ public class OperationalReportServiceImpl implements OperationalReportService {
                 response.setResponseCode(HttpStatus.BAD_REQUEST);
                 response.getParams().setStatus(Constants.FAILED);
                 response.getParams().setErrmsg("User Id does not exist");
+                return response;
             }
             Map<String, Map<String, String>> userInfoMap = new HashMap<>();
             userUtilityService.getUserDetailsFromDB(
-                    Collections.singletonList(userId), Arrays.asList("rootOrgId", "userId"), userInfoMap);
+                    Collections.singletonList(userId), Arrays.asList(Constants.ROOT_ORG_ID, Constants.USER_ID), userInfoMap);
             Map<String, String> userDetailsMap = userInfoMap.get(userId);
-            String rootOrg = userDetailsMap.get("rootOrgId");
+            String rootOrg = userDetailsMap.get(Constants.ROOT_ORG_ID);
             String objectKey = serverProperties.getOperationalReportFolderName() + "/" + rootOrg + "/" + serverProperties.getOperationReportFileName();
             logger.info("Object key for the operational Reports : " + objectKey);
             Model.Blob blob = storageService.getObject(serverProperties.getOperationalReportFolderName(), objectKey, Option.apply(Boolean.FALSE));
             if (blob != null) {
                 logger.info("File details" + blob.lastModified());
                 logger.info("File details" + blob.metadata());
-                response.put("lastModified", blob.lastModified());
-                response.put("fileMetaData", blob.metadata());
+                response.put(Constants.LAST_MODIFIED, blob.lastModified());
+                response.put(Constants.FILE_METADATA, blob.metadata());
             }
         } catch (Exception e) {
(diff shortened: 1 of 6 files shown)


src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -367,6 +372,30 @@ private long getParsedDate(String reportExpiryDate) throws ParseException {
         return reportExpiryDateMillis;
     }
 
+
+    public SBApiResponse getFileInfo(String authToken) {
+        logger.info("Inside the getFileInfo()");
+        String userId = accessTokenValidator.fetchUserIdFromAccessToken(authToken);
+        if (null == userId) {
+            throw new HttpServerErrorException(HttpStatus.INTERNAL_SERVER_ERROR, "UserId Couldn't fetch from auth token");
```
karthik-tarento: this API must return response object with proper error details instead of throwing exception..

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -367,6 +372,30 @@ private long getParsedDate(String reportExpiryDate) throws ParseException {
         return reportExpiryDateMillis;
     }
 
+
+    public SBApiResponse getFileInfo(String authToken) {
```
karthik-tarento, line 376: should have try catch and from catch block should update response object with 500 error details

karthik-tarento (changes requested):

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
tarentomaheshvakkund: have created a response object and returned.

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
tarentomaheshvakkund: have created a try catch block and returned the internal server error response in the catch block

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -367,6 +372,39 @@ private long getParsedDate(String reportExpiryDate) throws ParseException {
...
+        SBApiResponse response = ProjectUtil.createDefaultResponse(Constants.GET_FILE_INFO_OPERATIONAL_REPORTS);
+        try {
+            logger.info("Inside the getFileInfo()");
+            String userId = accessTokenValidator.fetchUserIdFromAccessToken(authToken);
+            if (null == userId) {
+                response.setResponseCode(HttpStatus.BAD_REQUEST);
+                response.getParams().setStatus(Constants.FAILED);
+                response.getParams().setErrmsg("User Id does not exist");
```
karthik-tarento, line 384: return the response immediately in case of error.

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -367,6 +372,39 @@ private long getParsedDate(String reportExpiryDate) throws ParseException {
...
+            }
+            Map<String, Map<String, String>> userInfoMap = new HashMap<>();
+            userUtilityService.getUserDetailsFromDB(
+                    Collections.singletonList(userId), Arrays.asList("rootOrgId", "userId"), userInfoMap);
+            Map<String, String> userDetailsMap = userInfoMap.get(userId);
+            String rootOrg = userDetailsMap.get("rootOrgId");
```
karthik-tarento: use the existing constants instead of rootOrgId string

karthik-tarento (changes requested):

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
tarentomaheshvakkund: returning the response object.

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
tarentomaheshvakkund: using the constants value.

PR #494, opened 2024-03-08
KB-3171 | BE | Dev | Enhancement of the API for downloading a password-protected zip report

tarentomaheshvakkund:
1. Enhancements in the Download file API to download the file based on the authtoken.

### DIFF
--- a/src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
+++ b/src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
@@ -168,42 +168,40 @@ public class OperationalReportServiceImpl implements OperationalReportService {
     }
 
 
-    /**
-     * Downloads a file from storage, processes it, and prepares it for download.
-     *
-     * @param reportType The type of report.
-     * @param date       The date of the report.
-     * @param orgId      The organization ID.
-     * @param fileName   The name of the file to be downloaded.
-     * @return A ResponseEntity containing the file for download.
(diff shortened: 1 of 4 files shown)


src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
```diff
@@ -168,42 +168,40 @@ public void init() {
...
     @Override
-    public ResponseEntity<InputStreamResource> downloadFile(String reportType, String date, String orgId, String fileName) throws IOException {
+    public ResponseEntity<InputStreamResource> downloadFile(String authToken) {
         HttpHeaders headers = new HttpHeaders();
         try {
-            // Construct the object key for downloading from storage
-            String objectKey = serverProperties.getReportDownloadFolderName() + "/" + reportType + "/" + date + "/" + orgId + "/" + fileName;
+            String userId = accessTokenValidator.fetchUserIdFromAccessToken(authToken);
+            if (null == userId) {
+                ResponseEntity.status(HttpStatus.BAD_REQUEST).body("User Id does not exist");
```
karthik-tarento, line 177: Should return the responseEntity from here

karthik-tarento (changes requested):

src/main/java/org/sunbird/operationalreports/service/OperationalReportServiceImpl.java
karthik-tarento: Updating it myself as this needs to be deployed ASAP

PR #500, opened 2024-03-13
Adding authToken validation in bulkUpload

Sahil-tarento:


src/main/java/org/sunbird/user/registration/model/UserRegistration.java
```diff
@@ -101,6 +102,14 @@ public void setProposedDeptName(String proposedDeptName) {
...
 	}
 
+	public String getUserAuthToken() {
+		return userAuthToken;
+	}
+
+	public void setUserAuthToken(String userAuthToken) {
+		this.userAuthToken = userAuthToken;
+	}
+
```
karthik-tarento, line 112: Did we cross check the usage of userRegistration class ?
is there any side effect due to this change ?

Do test the user registration feature once this is deployed in PreProd

src/main/java/org/sunbird/user/service/UserUtilityServiceImpl.java
```diff
@@ -589,9 +589,11 @@ public String createBulkUploadUser(UserRegistration userRegistration) {
 		requestBody.put(Constants.PHONE_VERIFIED, true);
 		requestBody.put(Constants.ROLES, Arrays.asList(Constants.PUBLIC));
 		request.put(Constants.REQUEST, requestBody);
+		Map<String, String> headerValues = ProjectUtil.getDefaultHeaders();
+		headerValues.put(Constants.X_AUTH_TOKEN, userRegistration.getUserAuthToken());
```
karthik-tarento: Need to add this header only if the value is available.
getDefaultHeaders adds content-type header -- is this added again ?

karthik-tarento (changes requested):

src/main/java/org/sunbird/user/registration/model/UserRegistration.java
Sahil-tarento: @JsonIgnoreProperties(ignoreUnknown = true) is added on the entity class

src/main/java/org/sunbird/user/service/UserUtilityServiceImpl.java
Sahil-tarento: added the empty check

karthik-tarento (approved):

PR #519, opened 2024-03-26
Fix bulkUpload and adding the key value

Sahil-tarento:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -251,16 +255,22 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
                         if (invalidErrList.isEmpty()) {
-                            userRegistration.setUserAuthToken(inputDataMap.get(Constants.X_AUTH_TOKEN));
-                            String responseCode = userUtilityService.createBulkUploadUser(userRegistration);
-                            if (!responseCode.equalsIgnoreCase(Constants.OK)) {
-                                failedRecordsCount++;
-                                statusCell.setCellValue(Constants.FAILED_UPPERCASE);
-                                errorDetails.setCellValue(responseCode);
+                            if (StringUtils.isBlank(redisCacheMgr.getCache(userRegistration.getPhone()+ "" +userRegistration.getEmail()))) {
+                                redisCacheMgr.putCache(userRegistration.getPhone().trim()+ "" +userRegistration.getEmail(),
+                                        userRegistration);
```
karthik-tarento: No need to put the data -- only key is sufficient -- just check the key exists

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -251,16 +255,22 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
+                                userRegistration.setUserAuthToken(inputDataMap.get(Constants.X_AUTH_TOKEN));
+                                String responseCode = userUtilityService.createBulkUploadUser(userRegistration);
+                                if (!responseCode.equalsIgnoreCase(Constants.OK)) {
```
karthik-tarento: always check against constant

karthik-tarento (changes requested):

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
Sahil-tarento: Done

src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
Sahil-tarento: Done

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -771,7 +771,7 @@ public SBApiResponse bulkUpload(MultipartFile mFile, String orgId, String channe
 			response.getResult().putAll(uploadedFile);
 			uploadedFile.put(Constants.ORG_NAME, channel);
 			uploadedFile.put(Constants.X_AUTH_TOKEN, userAuthToken);
-			kafkaProducer.push(serverConfig.getUserBulkUploadTopic(), uploadedFile);
+			kafkaProducer.pushWithKey(serverConfig.getUserBulkUploadTopic(), uploadedFile, orgId);
```
karthik-tarento, line 774: What is the use of this key -- if it is not utilised at the consumption ?
The idea should be:
use the key value at the consumption... 
push the file with specific key value --> ex -> bulkUploadFile
also push the user create request to in the kafka with key --> ex -> userCreateReq

At the consumer -- check the key value and process the record accordingly...

karthik-tarento (changes requested):

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
Sahil-tarento: Using CompletableFuture.runAsync() in your consumer means 
that message processing can happen asynchronously and potentially 
concurrently across multiple threads. While messages within the same 
partition and with the same key will be processed sequentially, messages
from different partitions or with different keys may be processed 
concurrently.

PR #524, opened 2024-04-03
Bulk upload changes and email and group validation changes

Sahil-tarento:


src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
```diff
@@ -1930,6 +1934,26 @@ public List<String> adminApprovalFields() {
...
+
+	private boolean isFileExistForProcessingForMDO(String mdoId) {
+		Map<String, Object> bulkUplaodPrimaryKey = new HashMap<String, Object>();
+		bulkUplaodPrimaryKey.put(Constants.ROOT_ORG_ID, mdoId);
+		List<String> fields = Arrays.asList(Constants.ROOT_ORG_ID, Constants.IDENTIFIER, Constants.STATUS);
+
+		List<Map<String, Object>> bulkUploadMdoList = cassandraOperation.getRecordsByPropertiesWithoutFiltering(
+				Constants.KEYSPACE_SUNBIRD, Constants.TABLE_USER_BULK_UPLOAD, bulkUplaodPrimaryKey, fields);
+		if (CollectionUtils.isEmpty(bulkUploadMdoList)) {
+			return true;
```
karthik-tarento: Should be false

karthik-tarento (changes requested):

src/main/java/org/sunbird/profile/service/ProfileServiceImpl.java
Sahil-tarento: Done

PR #530, opened 2024-04-16
4.8.13 calendar v1

Sahil-tarento:


src/main/java/org/sunbird/calendar/controller/CalendarBulkUploadController.java
SaipradeepR, line 1: Could you create a variable to store the value of URLDecoder.decode(channel, "UTF-8") and utilize it across multiple locations?

SaipradeepR: Ensure that the controller handle does not throw UnsupportedEncodingException and instead, provide a readable error message.

src/main/java/org/sunbird/calendar/service/CalendarBulkUploadConsumer.java
SaipradeepR, line 1: Could you make the log logger.error("Error in Calendar Bulk Upload Consumer: Invalid Kafka Msg") dynamic so that it provides details about which record failed in multi thread processing ?

SaipradeepR: Can you use a literal topic instead of topics in the '${kafka.topics.calendar.bulk.upload.event.group}'?

src/main/java/org/sunbird/calendar/service/EventUtilityService.java
SaipradeepR, line 1: Create constant for "/"

Sahil-tarento: Done

src/main/java/org/sunbird/calendar/service/CalendarBulkUploadConsumer.java
Sahil-tarento: In this method 
initiateCalendarEventBulkUploadProcess
we already have validateKafkaMsg will do this

Sahil-tarento: This is property name and everwhere it is used in this way only 
Can you use a literal topic instead of topics in the '${kafka.topics.calendar.bulk.upload.event.group}'?
sai and the grp name is calendarBulkUpload
follow the pattern.

src/main/java/org/sunbird/calendar/controller/CalendarBulkUploadController.java
Sahil-tarento: Done

PR #534, opened 2024-04-18
recommend content api changes for the gyan karmayogi

shankaragoudab:
changes made in the construction of url


src/main/java/org/sunbird/user/service/UserUtilityServiceImpl.java
```diff
@@ -827,10 +827,15 @@ public SBApiResponse recommendContent(String authUserToken, Map<String, Object>
 					emailResponseList.add(email);
 				}
 			}
+			logger.info("emailist : " +  emailResponseList);
```
karthik-tarento, line 830: No need to log this detail

karthik-tarento (changes requested):

PR #540, opened 2024-04-25
KAR-548 - Enhancement on getUserProgress API to paginate response

sreeragksgh:


src/main/java/org/sunbird/cassandra/utils/CassandraOperationImpl.java
```diff
@@ -449,5 +449,103 @@ public Long getRecordCountWithUserId(String keyspace, String tableName, String u
...
+				response.put(Constants.PAGE_ID, results.getExecutionInfo().getPagingState().toString());
+			}
+		} catch (Exception e) {
+			logger.error(Constants.EXCEPTION_MSG_FETCH + tableName + " : " + e.getMessage(), e);
+		}
+		return response;
+	}
+
+	@Override
+	public List<Map<String,Object>> getCountOfRecordByIdentifier(String keyspaceName, String tableName, Map<String,Object> key, String field) {
```
karthik-tarento: Need rethink this method and find an optimised way.
Find what is the better way to get count from cassandra table and return only the count instead of list of objects.

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -218,6 +222,78 @@ public Map<String, Object> getUserProgress(SunbirdApiRequest requestBody, String
...
+		}
+		Integer limit = (Integer) request.get(Constants.LIMIT);
+		if (limit == null) {
+			limit = 10; // check default limit add it as constants
+		}
+		Integer currentOffSetFromRequest = (Integer) request.get(Constants.OFFSET);
+		if (currentOffSetFromRequest == null) {
+			currentOffSetFromRequest = 0;
+		}
+		String pageId = (String) request.get(Constants.PAGE_ID);
```
karthik-tarento, line 243: This request object is created newly in getUserProgress method and it doesn't pass the pageId value... 
PageId must be sent back to UI and UI should return the value for next page read... 
If pageId is not in request body of which UI sent -- then backend should return the 1st page user details...

karthik-tarento (changes requested):

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -150,6 +151,8 @@ public void enrichProgressDetails(String authUserToken, MandatoryContentResponse
 
 	public Map<String, Object> getUserProgress(SunbirdApiRequest requestBody, String authUserToken, String rootOrgId, String userChannel) {
```
karthik-tarento, line 152: the existing API seems like returning all batches from course if possible... 
if course has 3 batches and each batch has 5 users -- this API would return 15 users details... 

the new design tries to get users from specific course batch -- and tries to provide pagination... 
do add a v2/progress/getUserProgressByBatch method and copy this updated implementation to new method... 

do get rid of the for loop for the batch in new method. make sure validation done for all newly added fields from requestBody....

karthik-tarento (changes requested):

PR #543, opened 2024-04-26
4.8.12 microsites

dkttarento:
Removed unnecessary properties &  fixed org array order


src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+               response.put(MESSAGE, "Organization is Missing");
+               response.setResponseCode(HttpStatus.BAD_REQUEST);
+               return response;
+           }
+           Map<String, String> payloadToRedisKeyMapping = serverProperties.getPayloadToRedisKeyMapping();
+           contextTypeList.addAll(payloadToRedisKeyMapping.keySet());
+           Map<String, Object> aggregateData = new HashMap<>();
+           List<String> topContentRedis;
+           List<String> limitCourses;
+           int limit = Optional.ofNullable(request.get(LIMIT)).map(l -> (Integer) l).orElse(0);
```
Sahil-tarento: It should be some default value for limit

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
         return "";
     }
 
-    public Map<String, Object> trendingContentSearch(Map<String, Object> requestBody, String token) throws Exception {
+    public SBApiResponse trendingContentSearch(Map<String, Object> requestBody, String token) throws Exception {
+       SBApiResponse response = ProjectUtil.createDefaultResponse(Constants.MICROSITE_TOP_FEATURE_CONTENT_API);
+       try {
+           HashMap<String, Object> request = requestBody.containsKey(REQUEST) ? (HashMap<String, Object>) requestBody.get(REQUEST) : new HashMap<>();
```
Sahil-tarento: Use the Generic Interface Impl for Map<String, Object>
Instead of creating new Object new HashMap<>

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+           if (request.isEmpty()) {
```
Sahil-tarento: Just used MapUtils and remove new HashMap<>

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+               response.getParams().setStatus(Constants.FAILED);
+               response.put(MESSAGE, "Request is Missing");
+               response.setResponseCode(HttpStatus.BAD_REQUEST);
+               return response;
+           }
+           HashMap<String, Object> filter = request.get(FILTERS) == null ? new HashMap<>() : ((HashMap<String, Object>) request.get(FILTERS));
```
Sahil-tarento: Use the Generic Interface Impl for Map<String, Object>
Instead of creating new Object new HashMap<>

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+           if (filter.isEmpty()) {
```
Sahil-tarento: Just used MapUtils and remove new HashMap<>

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+               response.getParams().setStatus(Constants.FAILED);
+               response.put(MESSAGE, "Filter is Missing");
+               response.setResponseCode(HttpStatus.BAD_REQUEST);
+               return response;
+           }
+           List<String> contextTypeList = new ArrayList<>();
```
SaipradeepR: Where we are filling in the contextTypes based on the payload? why u r checking not empty here to throw BAD request

Sahil-tarento: Need to add contextType list from request where is that impl

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+           }
+           String org = (filter).get(ORGANISATION) == null ? "" : ((String) (filter).get(ORGANISATION));
+           if (org.isEmpty()) {
+               response.getParams().setStatus(Constants.FAILED);
+               response.put(MESSAGE, "Organization is Missing");
+               response.setResponseCode(HttpStatus.BAD_REQUEST);
+               return response;
+           }
+           Map<String, String> payloadToRedisKeyMapping = serverProperties.getPayloadToRedisKeyMapping();
+           contextTypeList.addAll(payloadToRedisKeyMapping.keySet());
```
SaipradeepR: You have to read selected context Type from request payload not all contexts from properties.

Sahil-tarento: No need of this impl why we need that

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+                   aggregateData.put(contextType, limitCourses);
+               }
+           }
+           Map<String, Object> compositeSearchRes;
+           List<Map<String, Object>> contentList = new ArrayList<>();
+           Map<String, Object> resultMap;
+           if (!aggregateData.isEmpty()) {
+               List<String> compositeKeyList = new ArrayList<>(payloadToRedisKeyMapping.keySet());
+               for (int i = 0; i < aggregateData.size(); i++) {
+                   List<String> searchIds = (List<String>) aggregateData.get(compositeKeyList.get(i));
```
Sahil-tarento, line 242: Why you do composite search contentService.readContentFromCache for getting the value for content and add the same way i add for topReview.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+           }
+           Map<String, String> payloadToRedisKeyMapping = serverProperties.getPayloadToRedisKeyMapping();
+           contextTypeList.addAll(payloadToRedisKeyMapping.keySet());
+           Map<String, Object> aggregateData = new HashMap<>();
+           List<String> topContentRedis;
+           List<String> limitCourses;
+           int limit = Optional.ofNullable(request.get(LIMIT)).map(l -> (Integer) l).orElse(0);
+           for (String contextType : contextTypeList) {
+               String contextTypeValue = payloadToRedisKeyMapping.get(contextType);
+               topContentRedis = redisCacheMgr.hget(contextTypeValue, serverProperties.getRedisInsightIndex(), new String[]{org});
```
SaipradeepR: rename topContentRedis param to valid generic param value.

Sahil-tarento (changes requested):

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
```diff
@@ -178,84 +179,72 @@ private String getEndDateFormat() {
...
+               for (int i = 0; i < aggregateData.size(); i++) {
+                   List<String> searchIds = (List<String>) aggregateData.get(compositeKeyList.get(i));
+                   compositeSearchRes = compositeSearch(searchIds, token);
+                   if (null == compositeSearchRes)
+                       compositeSearchRes = new HashMap<>();
+                   resultMap = compositeSearchRes.get(RESULT) == null ? new HashMap<>() : (Map<String, Object>) compositeSearchRes.get(RESULT);
+                   contentList = resultMap.get(CONTENT) == null ? new ArrayList<>() : (List<Map<String, Object>>) resultMap.get(CONTENT);
+               }
+           }
+           response.put(RESPONSE, contentList);
```
SaipradeepR: It will only contain the content list of the last contextType because within the loop above, contentList is consistently being replaced.

src/main/java/org/sunbird/trending/controller/TrendingController.java
```diff
@@ -21,12 +22,12 @@ public ResponseEntity<Map<String, Object>> trendingSearch(
         return new ResponseEntity<>(compositeSearchRes, HttpStatus.OK);
     }
 
-    @PostMapping("admin/trending/content/search")
-    public ResponseEntity<Map<String, Object>> trendingContentSearch(
+    @PostMapping("/microsite/topFeatured/content/search")
+    public ResponseEntity<SBApiResponse> trendingContentSearch(
```
SaipradeepR, line 26: Instead of using trendingContentSearch, derive a more generic name following the discussion with Sahil.

src/main/java/org/sunbird/trending/controller/TrendingController.java
```diff
@@ -21,12 +22,12 @@ public ResponseEntity<Map<String, Object>> trendingSearch(
         return new ResponseEntity<>(compositeSearchRes, HttpStatus.OK);
     }
 
-    @PostMapping("admin/trending/content/search")
-    public ResponseEntity<Map<String, Object>> trendingContentSearch(
+    @PostMapping("/microsite/topFeatured/content/search")
```
SaipradeepR: derive generic url as similar like content/context/search

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: @SaipradeepR , so the response should be based only on "contextType" only ? In that case why we are having "contextType" values in properties file ?

dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested

src/main/java/org/sunbird/trending/controller/TrendingController.java
dkttarento: As discussed we are using , "content/aggregation/search"

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Not clear.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: Implimented as suggested.

src/main/java/org/sunbird/trending/controller/TrendingServiceImpl.java
dkttarento: value of limit is picked from request, if limit is missing i am returning a validation error.

SaipradeepR:
I've noticed there are still some issues. Please address these during integration testing to allow the UI merging to proceed for now.

PR #544, opened 2024-04-29
KAR-548 - NEW V2 API for getUserProgress with Pagination

sreeragksgh:


src/main/java/org/sunbird/cassandra/utils/CassandraOperationImpl.java
```diff
@@ -449,5 +449,107 @@ public Long getRecordCountWithUserId(String keyspace, String tableName, String u
...
+			if (MapUtils.isEmpty(key)) {
+				throw new IllegalArgumentException("Key parameter cannot be null");
+			}
+			Session session = connectionManager.getSession(keyspaceName);
+			Builder selectBuilder;
+			if (StringUtils.isNotEmpty(field)) {
+				selectBuilder = QueryBuilder.select().count(field);
+			} else {
+				selectBuilder = QueryBuilder.select().countAll();
+			}
```
karthik-tarento: count should be applied for all 
just check field is not empty -- if empty throw error

src/main/java/org/sunbird/cassandra/utils/CassandraOperationImpl.java
```diff
@@ -449,5 +449,107 @@ public Long getRecordCountWithUserId(String keyspace, String tableName, String u
...
+						}
+					} else {
+						Clause clause = QueryBuilder.eq(entry.getKey(), entry.getValue());
+						selectWhere.and(clause);
+					}
+				}
+			}
+			ResultSet results = session.execute(selectQuery);
+			response = CassandraUtil.createResponse(results);
+			count = ((Long)((Map<String,Object>)response.get(0)).get("system.count(" + field.toLowerCase() + ")"));
```
karthik-tarento, line 543: since you are utilizing field value - it must not be null or empty

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -218,6 +217,77 @@ public Map<String, Object> getUserProgress(SunbirdApiRequest requestBody, String
...
+			propertyMap.put(Constants.COURSE_ID, request.get(Constants.COURSE_ID));
+			userEnrolmentList.addAll(cassandraOperation.getRecordsByPropertiesWithoutFiltering(Constants.KEYSPACE_SUNBIRD_COURSES,
+					Constants.TABLE_USER_ENROLMENT, propertyMap,
+					Arrays.asList(Constants.USER_ID_CONSTANT, Constants.COURSE_ID,
+							Constants.BATCH_ID, Constants.COMPLETION_PERCENTAGE, Constants.PROGRESS,
+							Constants.STATUS, Constants.ISSUED_CERTIFICATES), cbExtServerProperties.getBatchEnrolmentReturnSize()));
+			// restricting with only 100 items in the response
+			if (userEnrolmentList.size() > cbExtServerProperties.getBatchEnrolmentReturnSize()) {
+				userEnrolmentList = userEnrolmentList.subList(0, cbExtServerProperties.getBatchEnrolmentReturnSize());
+			}
```
karthik-tarento: Since its driven by limit -- we just need to make sure limit is not greater than cbExtServerProperties.getBatchEnrolmentReturnSize() -- if so -- need to return error ??

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -312,4 +382,109 @@ private int getLeafCountForTheCourse(String rootOrgId, String courseId, String u
...
+		Map<String, Object> result = new HashMap<String, Object>();
+		List<String> userList = new ArrayList<String>();
+		Boolean active = (Boolean) request.get(Constants.ACTIVE);
+		if (null == active) {
+			active = true;
+		}
+		Integer limit = (Integer) request.get(Constants.LIMIT);
+		if (limit == null) {
+			limit = 10; // check default limit add it as constants
+		}
```
karthik-tarento: limit can't be null right ?
it's already checked inside validation method?

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -312,4 +382,109 @@ private int getLeafCountForTheCourse(String rootOrgId, String courseId, String u
...
+		Integer currentOffSetFromRequest = (Integer) request.get(Constants.OFFSET);
+		if (currentOffSetFromRequest == null) {
+			currentOffSetFromRequest = 0;
+		}
```
karthik-tarento: offset can't be null as it is checked in validation method

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -312,4 +382,109 @@ private int getLeafCountForTheCourse(String rootOrgId, String courseId, String u
...
+			//We may have read the given limit... if so, break from while loop
+			if (userList.size() == limit) {
+				break;
+			}
+		} while (org.codehaus.plexus.util.StringUtils.isNotBlank(currentPagingState));
+		if (org.codehaus.plexus.util.StringUtils.isNotBlank(previousPageId)) {
+			result.put(Constants.PAGE_ID, previousPageId);
+			if (currentOffSet >= limit) {
+				currentOffSet = currentOffSet - limit;
+			}
```
karthik-tarento, line 484: why are we doing this?

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -312,4 +382,109 @@ private int getLeafCountForTheCourse(String rootOrgId, String courseId, String u
...
+				previousPageId = null;
+				break;
+			}
+			for (Map<String, Object> userCourse : userCoursesList) {
+				//From this page, we have already read some records, so skip the records
+				if (currentOffSetFromRequest > 0) {
+					currentOffSetFromRequest--;
+					continue;
+				}
+				if (userCourse.get(Constants.ACTIVE) != null
```
karthik-tarento, line 462: Do add an else part for this condition... 
inside else - just log an error statement with userId, courseId and batchId details...

src/main/java/org/sunbird/progress/service/MandatoryContentServiceImpl.java
```diff
@@ -312,4 +382,109 @@ private int getLeafCountForTheCourse(String rootOrgId, String courseId, String u
...
+						previousPageId = pageId != null ? pageId : previousPageId;
+						break;
+					}
+				}
+				currentOffSet++;
+			}
+			//We may have read the given limit... if so, break from while loop
+			if (userList.size() == limit) {
+				break;
+			}
```
karthik-tarento, line 478: userList is updated only in line #466 - so checking the size against limit after that is fine... 

but here -- will this be ever reach this place?

karthik-tarento (changes requested):

PR #550, opened 2024-05-09
Org Upload Assests

Sahil-tarento:


src/main/java/org/sunbird/storage/controller/StorageController.java
```diff
@@ -71,4 +71,12 @@ public ResponseEntity<?> getFileInfoSPV(@RequestHeader(Constants.X_AUTH_TOKEN) S
 											@PathVariable("date") String date) {
 		return storageService.getFileInfoSpv(userToken, date);
 	}
+
+	@PostMapping("/orgStoreUpload")
+	public ResponseEntity<?> orgStoreUpload(@RequestHeader(Constants.X_AUTH_USER_ORG_ID) String orgId
```
karthik-tarento: Let's read the orgId from userToken...
This way user org would be properly identified instead of accepting header values.

karthik-tarento (changes requested):

src/main/java/org/sunbird/storage/controller/StorageController.java
Sahil-tarento: Done

PR #573, opened 2024-05-27
Bug Fix for validation

Sahil-tarento:


src/main/java/org/sunbird/common/util/ProjectUtil.java
```diff
@@ -202,4 +202,12 @@ public static Boolean validateDate(String dateString){
 	public static Boolean validateEmployeeId(String employeeId) {
 		return employeeId.matches("^(?=.*\\d|[a-zA-Z]{30})[a-zA-Z0-9 .-]{1,30}$"); // Allow only alphanumeric, numeric and restrict if only alphabets character
 	}
+
+	public static Boolean validateRegexPatternWithNoSpecialCharacter(String regex) {
+		return regex.matches("^[a-zA-Z0-9 -()]*$");
+	}
+
+	public static Boolean validatePinCode(String regex) {
+		return regex.matches("^[0-9]{4,7}$");
```
karthik-tarento: why length is 4 to 7?
do we have pincode in this range?

karthik-tarento (changes requested):

src/main/java/org/sunbird/common/util/ProjectUtil.java
Sahil-tarento: No @karthik-tarento  Updated to 6 digit standard used in india

PR #577, opened 2024-05-29
Adding logs to analyse OTP generation API 5xx issue in production

sreeragksgh:


src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
```diff
@@ -195,10 +195,12 @@ public SBApiResponse getDeptDetails() {
 	}
 
 	public SBApiResponse generateOTP(Map<String, Object> otpRequests) {
+		LOGGER.info("inside generateOTP method");
```
karthik-tarento: This is not required

src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
```diff
@@ -195,10 +195,12 @@ public SBApiResponse getDeptDetails() {
...
 		SBApiResponse response = createDefaultResponse(Constants.USER_GENERATE_OTP);
 		String errMsg = validateOTPPayload(otpRequests);
 		if (StringUtils.isBlank(errMsg)) {
 			try {
+				LOGGER.info("validation on OTP Request is successful");
```
karthik-tarento: this is not required
Do add the log if API call is successful or failed

src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
```diff
@@ -218,8 +220,11 @@ public SBApiResponse generateOTP(Map<String, Object> otpRequests) {
 				LOGGER.error(String.format("Exception in %s : %s", "generateOTP", e.getMessage()), e);
 				errMsg = "Failed to process message. Exception: " + e.getMessage();
 			}
+		}else {
+			LOGGER.error("Validation failed for OTP generation request, error message : ",errMsg);
```
karthik-tarento: this is not required

karthik-tarento (changes requested):

PR #579, opened 2024-05-30
ResponseCode correction cb-ext generate ext OTP API

sreeragksgh:


src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
```diff
@@ -212,6 +212,13 @@ public SBApiResponse generateOTP(Map<String, Object> otpRequests) {
 					response.getParams().setMsgid(UUID.randomUUID().toString());
 					response.setResult(new HashMap<>());
 					response.getResult().put(Constants.RESPONSE, Constants.SUCCESS.toUpperCase());
+				} else if (Constants.TOO_MANY_REQUESTS.equalsIgnoreCase((String) apiResponse.get(Constants.RESPONSE_CODE))) {
```
karthik-tarento: Need to handle the error message as it is -- instead of handling one by one.
what will happen if it is 400 ??

karthik-tarento (changes requested):

PR #608, opened 2024-06-17
4.8.15 dev v2

anilkumarkammalapalli:


src/main/java/org/sunbird/profile/service/UserBulkUploadService.java
```diff
@@ -417,6 +423,273 @@ private void processBulkUpload(HashMap<String, String> inputDataMap) throws IOEx
...
+                        if (!ProjectUtil.validateExternalSystem(userRegistration.getExternalSystem())) {
+                            invalidErrList.add("Invalid External System Name : External System Name can contain only alphabets and alphanumeric and can have a max length of 255");
+                        }
+                    }
+
+                    if (!isFieldEmpty(record, 13)) {
+                        String tagStr = record.get(13).trim();
+                        List<String> tagList = new ArrayList<>();
+                        if (!StringUtils.isEmpty(tagStr)) {
+                            String[] tagStrList = tagStr.split(",", -1);
```
karthik-tarento: Let's expect TAG value to be given like - Test1 & Test2
Split the TAG using "&" and trim it...

karthik-tarento (changes requested):

PR #641, opened 2024-07-11
New API to Fetch WhiteListed Domains for email - KB-5850

sreeragksgh:


sreeragksgh:
Jira Ticket : https://karmayogibharat.atlassian.net/browse/KB-5850

src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
```diff
@@ -588,4 +621,29 @@ private Boolean isApprovedDomains(String emailDomain, String domainType) {
 				Constants.KEYSPACE_SUNBIRD, Constants.TABLE_MASTER_DATA, propertyMap, Arrays.asList(Constants.CONTEXT_TYPE, Constants.CONTEXT_NAME));
 		return CollectionUtils.isNotEmpty(listOfDomains);
 	}
+
+	private List<String> getApprovedDomainsFromDB() {
```
karthik-tarento: what will happen if the domain exist in both list?

karthik-tarento (changes requested):

src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
sreeragksgh: Updated code to tackle this

src/main/java/org/sunbird/user/registration/service/UserRegistrationServiceImpl.java
```diff
@@ -305,6 +306,38 @@ public void initiateCreateUserFlow(String registrationCode) {
 		}
 	}
 
+	public SBApiResponse getApprovedDomains() {
+		SBApiResponse response = createDefaultResponse(Constants.API_APPROVED_DOMAINS);
```
karthik-tarento: I thought this method adds all the necessary information into the response... 

if not -- just use ProjectUtil.createDefaultResponse

karthik-tarento (changes requested):

PR #642, opened 2024-07-12
4.8.15 dev v4

anilkumarkammalapalli:


src/main/java/org/sunbird/catalog/service/CatalogServiceImpl.java
```diff
@@ -432,4 +432,153 @@ private String validateSubSectorCreateReq(Map<String, Object> sectorRequest) {
...
+			return errMsg;
+		}
+		List<String> missingParams = new ArrayList<>();
+		String name = (String) designationRequest.get(Constants.NAME);
+		if (StringUtils.isBlank(name)) {
+			missingParams.add(Constants.NAME);
+		}
+		String refId = (String) designationRequest.get(Constants.REF_ID);
+		if (StringUtils.isBlank(refId)) {
+			missingParams.add(refId);
```
karthik-tarento: suppose to add Constants.REF_ID

src/main/java/org/sunbird/catalog/service/CatalogServiceImpl.java
```diff
@@ -432,4 +432,153 @@ private String validateSubSectorCreateReq(Map<String, Object> sectorRequest) {
...
+		}
+		String refType = (String) designationRequest.get(Constants.REF_TYPE);
+		if (StringUtils.isBlank(refType)) {
+			missingParams.add(refType);
```
karthik-tarento: suppose to add Constants.REF_TYPE

src/main/java/org/sunbird/catalog/service/CatalogServiceImpl.java
```diff
@@ -432,4 +432,153 @@ private String validateSubSectorCreateReq(Map<String, Object> sectorRequest) {
...
+			response.getParams().setStatus(Constants.FAILED);
+			response.setResponseCode(HttpStatus.INTERNAL_SERVER_ERROR);
+		}
+		return response;
+	}
+
+	private Map<String, Object> createRequestBody(String id, String designation, List<String> refNodes) {
+		try {
+			Map<String, Object> reqBodyMap = new HashMap<>();
+			reqBodyMap.put(Constants.ID, id);
```
karthik-tarento: why REF_ID and REF_TYPE are not used here ?
If those are not required then why those are mandatory?

src/main/java/org/sunbird/catalog/service/CatalogServiceImpl.java
```diff
@@ -432,4 +432,153 @@ private String validateSubSectorCreateReq(Map<String, Object> sectorRequest) {
...
+					.append(extServerProperties.getOdcsCategoryName());
+			Map<String, Object> termResponse = outboundRequestHandlerServiceImpl.fetchResultUsingPost(strUrl.toString(),
+					createReq, null);
+			if (termResponse != null
+					&& Constants.OK.equalsIgnoreCase((String) termResponse.get(Constants.RESPONSE_CODE))) {
+				Map<String, Object> resultMap = (Map<String, Object>) termResponse.get(Constants.RESULT);
+				List<String> termIdentifier = (List<String>) resultMap.getOrDefault(Constants.NODE_ID, "");
+				log.info("Created Designation successfully with name: " + name);
+				log.info("termIdentifier : " + termIdentifier);
+				response = updateDesignation(refId, name, termIdentifier);
```
karthik-tarento: what will happen if the updateDesignation fails ??

src/main/java/org/sunbird/catalog/controller/CatalogController.java
```diff
@@ -46,4 +46,11 @@ public ResponseEntity<SBApiResponse> createSubSector(@RequestBody Map<String, Ob
 		SBApiResponse response = catalogService.createSubSector(request);
 		return new ResponseEntity<>(response, response.getResponseCode());
 	}
+
+	@PostMapping("/designation/create")
```
karthik-tarento: this implementation is in cb-ext --- just to reuse the framework capability -- but nothing is reused from existing code.... 

after creating term object - you are still calling pores service to update designation... 

instead - just update the endpoints (or add new one) - in pores service to create / update -- that should create a term object - if term creation fails it returns... if term is created then updates the db...

karthik-tarento (changes requested):

src/main/java/org/sunbird/catalog/service/CatalogServiceImpl.java
anilkumarkammalapalli: id is ref_id only here. we are passing from caller function.

